PR #12, opened 2015-12-04
Users/drewta/delete files task lib

drewtabor:
updating for delete files task, adding L0s for rmRF

### DIFF
--- a/node/lib/vsotask.ts
+++ b/node/lib/vsotask.ts
@@ -1,10 +1,10 @@
 var Q = require('q');
 var shell = require('shelljs');
-var fs = require('fs');
 var path = require('path');
 var os = require('os');
 var minimatch = require('minimatch');
 var globm = require('glob');
+import fs = require('fs');
 import tcm = require('./taskcommand');
 import trm = require('./toolrunner');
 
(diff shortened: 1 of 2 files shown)


node/lib/vsotask.ts
```diff
@@ -219,6 +219,17 @@ export function getEndpointAuthorization(id: string, optional: boolean): Endpoin
...
 
 //-----------------------------------------------------
+// Fs Helpers
+//-----------------------------------------------------
+export interface FsStats extends fs.Stats {
+    
+}
+
+export function stats(path: string): FsStats {
+    return fs.statSync(path);
```
bryanmacfarlane: Add debug writes here.  debug the input.  debug write the properties and isDirectory, file and symbolic link.  Play around with debug accepting object or string and if object, it prints the directory properties if string, number ... ??  Thoughts

node/lib/vsotask.ts
```diff
@@ -317,6 +337,14 @@ export function find(findPath: string): string[] {
 export function rmRF(path: string): void {
     debug('rm -rf ' + path);
     shell.rm('-rf', path);
+    
+    var error: string = shell.error();
+    // if you try to delete a file that doesn't exist, desired result is achieved
+    // other errors are valid
+    if (error && !(error.indexOf('ENOENT') === 0)) {        
+        console.error(error)
+        exit(1);
```
bryanmacfarlane: I'm not sure we want exit code 1 communicated to the user.  That's fine if you're executing a tool.  

I think this needs to call setResult Fail with exit arg.  BTW, I think rmRF can take an optional arg continueOnError (which would be falsey if not suppled and exit).  But it gives task authors the ability to make clean up a best effort.

bryanmacfarlane:
I'm going to merge this, implement my feedback and then publish

PR #20, opened 2015-12-10
improve loc function to vso-task-lib.

TingluoHuang:


node/lib/vsotask.ts
```diff
@@ -103,44 +104,103 @@ export function exit(code: number): void {
...
 //-----------------------------------------------------
-var locStrings = {};
-var resourceFile;
+var locStringCache: {
+    [key: string]: string
+} = {};
+var resourceFile: string;
+var libResourceFileLoaded: boolean = false;
+
+function loadLocStrings(resourceFile: string): any {
```
scottdallamura: this can return { [key: string]: string; } instead of any

node/lib/vsotask.ts
```diff
@@ -385,26 +448,69 @@ export function mkdirP(p): boolean {
...
+                        return;
+                    }
+                });
+            }
 
-    debug(tool + '=' + toolPath);
-    return toolPath;
-}
+            // Command not found in Path, but the input itself is point to a file.
+            if (!toolPath && exist(tool) && stats(tool).isFile)
```
scottdallamura: I like brackets.
if (...) {
  ...
}

PR #25, opened 2016-01-15
Loc.

ericsciple:


node/gulpfile.js
```diff
@@ -34,8 +34,32 @@ gulp.task('clean', function (done) {
 	return del([buildRoot], done);
 });
 
-gulp.task('copy:manifest', ['clean'], function () {
-	return gulp.src(['package.json', 'lib.json', '../LICENSE', '../README.md'])
+gulp.task('loc:generate', ['clean'], function() {
+    // Build the content for the en-US resjson file.
+    var lib = require('./lib.json');
+    var strPath = './Strings/resources.resjson/en-US';
```
bryanmacfarlane: path.join('.', 'Strings', 'resources.resjson', 'en-US') is a better node way.  But you can change after PR - no reason to block anything

node/gulpfile.js
```diff
@@ -34,8 +34,32 @@ gulp.task('clean', function (done) {
...
+        for (var key in lib.messages) {
+            var messageKey = 'loc.messages.' + key;
+            strings[messageKey] = lib.messages[key];
+        }
+    }
+
+    // Create the en-US resjson file.
+    var enPath = path.join(strPath, 'resources.resjson');
+    var enContents = JSON.stringify(strings, null, 2);
+    fs.writeFile(enPath, enContents, function(err) {
```
bryanmacfarlane: This is an async function so you need to take done callback in function(){ above and call done(err).  Or, actually I would suggest just calling writeFileSync

powershell/gulpfile.js
```diff
@@ -25,7 +25,31 @@ gulp.task('clean', function (done) {
...
+        for (var key in lib.messages) {
+            var messageKey = 'loc.messages.' + key;
+            strings[messageKey] = lib.messages[key];
+        }
+    }
+
+    // Create the en-US resjson file.
+    var enPath = path.join(strPath, 'resources.resjson');
+    var enContents = JSON.stringify(strings, null, 2);
+    fs.writeFile(enPath, enContents, function(err) {
```
bryanmacfarlane: same

PR #27, opened 2016-01-25
Artifact logging command.

ericsciple:


powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
```diff
@@ -38,11 +38,21 @@ function Write-AssociateArtifact {
         [string]$Name,
         [Parameter(Mandatory = $true)]
         [string]$Path,
+        [Parameter(Mandatory = $true)]
+        [string]$Type,
```
ericsciple: @TingluoHuang It might be better to default the Type to 'FilePath', remove Mandatory = $true, add [ValidateNotNullOrEmpty()]. This would allow consumers to not specify the parameter at all (and default to FilePath) or specify a non-null/empty value. Otherwise the command would throw for null/empty. What ya think Ting?

ericsciple:
Reviewed offline w/ Ting.

PR #28, opened 2016-01-25
don't add quotes to path input.

TingluoHuang:


bryanmacfarlane:
Looks good

bryanmacfarlane:
I would bump the version

TingluoHuang:
Is the fix already in master?  We don't have releases branch.
What testing was done? Verify tasks that use this method continue works.
Does this fix change the AT/DT contract? No
Does this fix include any database servicing changes? No
Were tests added to cover this to make sure it doesn't regress?  We had a L0 test, but verify the wrong case, I change it to verify the correct scenario.

bryanmacfarlane:
We have release branches in tasks which is the ultimate vehicle of this fix.

bryanmacfarlane:
We should also state the customer problem

TingluoHuang:
I list the customer problem in the Bug.

TingluoHuang:
Bug 506238: vsts-task-lib getPathInput() always return path in quotes when path contain space

PR #30, opened 2016-01-27
when add path to toolrunner args, don't split string by space.

TingluoHuang:


node/lib/toolrunner.ts
```diff
@@ -92,10 +92,11 @@ export class ToolRunner extends events.EventEmitter {
      * Accepts a full string command line and a string array as well
      * Will handle double quoted args. E.g. '"arg one" two -z'
```
TingluoHuang: add comments.

node/lib/toolrunner.ts
```diff
@@ -105,12 +106,30 @@ export class ToolRunner extends events.EventEmitter {
...
+            else {
+                this._debug(this.toolPath + ' arg: ' + val);
+                this.args = this.args.concat(this._argStringToArray(val));    
+            }
         }
     }
 
     /**
+     * Add path argument
+     * Add path string to argument, path string should not contain double quoted
```
TingluoHuang: this just call arg(literal = true)

TingluoHuang:
Is the fix already in master?  N/A, we don't have release branch for task-lib
What testing was done? make sure the toolrunner is able to handle file path arg, do not split the string.
Does this fix change the AT/DT contract? N/A
Does this fix include any database servicing changes? N/A
Were tests added to cover this to make sure it doesn't regress?  we add L0 test to verify the scenario.

PR #31, opened 2016-01-29
finish node lib loc function.

TingluoHuang:


node/lib/task.ts
```diff
@@ -443,6 +472,7 @@ export function error(message: string): void {
 }
 
 export function debug(message: string): void {
+    console.log(message);
```
ericsciple: Is this left here by accident?

TingluoHuang: i am still pushing. :)

node/lib/task.ts
```diff
@@ -119,27 +119,55 @@ var locStringCache: {
 } = {};
 var resourceFile: string;
 var libResourceFileLoaded: boolean = false;
+var resourceCulture: string = 'en-US';
+
+function loadResJson(resjsonFile: string): any {
+    var resJson: {} = null;
+    if (exist(resjsonFile)) {
+        var resjsonContent = fs.readFileSync(resjsonFile, 'utf8').toString();
+        // remote BOM
```
ericsciple: fyi - typo: remote -> remove

node/lib/task.ts
```diff
@@ -119,27 +119,55 @@ var locStringCache: {
...
+        try {
+            resJson = JSON.parse(resjsonContent);
+        }
+        catch (err) {
+            debug('unable to parse resjson with err: ' + err.message);
+            resJson = null;
+        }
+    }
+    else {
+        debug('.resjson file not found: ' + resjsonFile);
```
ericsciple: It's your call, but one suggestion here: Most folks are not going to have a resjson file. I predict this "file not found" is going to confuse customers. Consider debug logging the file path if it is found, otherwise not debug log anything.

TingluoHuang: looks like i logged that pieces of information before i call the method. :)

TingluoHuang: I will try the user experience before complete PR.

node/lib/task.ts
```diff
@@ -119,27 +119,55 @@ var locStringCache: {
...
     if (exist(resourceFile)) {
-        debug('load loc strings from: ' + resourceFile);
+        debug('load strings from: ' + resourceFile);
         var resourceJson = require(resourceFile);
 
         if (resourceJson && resourceJson.hasOwnProperty('messages')) {
+            var locResourceJson = null;
+            // load up resource resjson for different culture
+            var localizedResourceFile = path.join(path.dirname(resourceFile), 'strings', 'resources.resjson', culture, 'resources.resjson');
+            if (exist(localizedResourceFile)) {
```
ericsciple: Not a big deal, but I noticed that you check `if exist` here and inside the function.

node/lib/task.ts
```diff
@@ -164,6 +192,7 @@ export function setResourcePath(path: string): void {
         resourceFile = null;
         libResourceFileLoaded = false;
         locStringCache = {};
+        resourceCulture = 'en-US';
```
ericsciple: Should this be `var resourceCulture = 'en-US';`?

node/lib/task.ts
```diff
@@ -171,9 +200,10 @@ export function setResourcePath(path: string): void {
         resourceFile = path;
         debug('set resource file to: ' + resourceFile);
 
-        var locStrs = loadLocStrings(resourceFile);
+        resourceCulture = getVariable('system.culture') || resourceCulture;
```
ericsciple: Will resourceCulture ever end up empty/null after this statement? If so, have you thought through the impact - i.e. will it error out later?

node/lib/task.ts
TingluoHuang: the resourceCulture is global variable, since the all L0 tests runs at the same process, i need cleanup those global vars.

node/lib/task.ts
```diff
@@ -356,7 +385,7 @@ export function getPathInput(name: string, required?: boolean, check?: boolean):
             checkPath(inval, name);
         }
     }
-    
+
     return inval;
 }
```
ericsciple: Looks good.

node/lib/task.ts
TingluoHuang: it should never be empty since resourceCulture init to be 'en-US'

ericsciple: It's initialized only if TASKLIB_INPROC_UNITS. I have no idea what that is, so I'll take your word :)

TingluoHuang: Line 122: var resourceCulture: string = 'en-US';

PR #36, opened 2016-02-04
Users/ersciple/m96loggingcmdsandtests

ericsciple:


powershell/Tests/L0/_suite.ts
```diff
@@ -0,0 +1,33 @@
...
+    before((done) => {
+        // init here
+        done();
+    });
+
+    after(function () {
+    });
+
+    if (ps) {
+        fs.readdirSync(__dirname).forEach(function (file: string) {
```
ericsciple: Here I'm just doing foreach ps1 file in the directory, run the script via psm.runPS( [...] ). This is just a convenience so we don't miss any files.

powershell/Tests/lib/TestHelpersModule/PrivateFunctions.ps1
```diff
@@ -1,9 +1,8 @@
 function Test-AreEqual {
```
ericsciple: The TestHelpersModule/** got an update from the copy under vso-agent-tasks.

powershell/Tests/lib/psRunner.ts
```diff
@@ -0,0 +1,79 @@
...
+            cmdLine,
+            {
+                cwd: wd,
+                // keep current env clean
+                env: process.env
+            },
+            (err, stdout, stderr) => {
+                if (stdout) {
+                    debug(stdout);
+                    if (stdout.search(/task\.complete.+result=failed/i) > 0)
```
ericsciple: This line is different from the runner under vso-agent-tasks. The reason is that the tests here are run through the module (Invoke-VstsTaskScript) so errors end up as task.complete result=failed over stdout.

The runner in the vso-agent-tasks repo considers the test failed if the exit code is non-zero.

powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
```diff
@@ -39,6 +39,23 @@ See https://github.com/Microsoft/vso-agent-tasks/blob/master/docs/authoring/comm
 .PARAMETER AsOutput
 Indicates whether to write the logging command directly to the host or to the output pipeline.
 #>
+function Write-AddBuildTag {
```
ericsciple: This is to support the logging commands changes Ting added to the agent.

powershell/definitions/Q.d.ts
```diff
@@ -0,0 +1,386 @@
+// Type definitions for Q
```
ericsciple: I just copied these definitions from the vso-agent-tasks repo. It seemed to be required to make "gulp test" and the psm.RunPS() work. I don't really know what I'm doing here, so just wanted to double-check this is correct thing to do.

powershell/gulpfile.js
```diff
@@ -2,12 +2,21 @@ var fs = require('fs');
 var path = require('path');
```
ericsciple: Gulpfile changes. I'm pretty sure I did this correctly, but wanted to make sure these new dependencies are OK.

powershell/package.json
```diff
@@ -24,7 +24,14 @@
   "devDependencies": {
     "del": "^2.2.0",
     "gulp": "^3.9.0",
+    "gulp-mocha": "^2.1.3",
```
ericsciple: Again, wanted to make sure these dependencies are OK.

PR #37, opened 2016-02-05
Updated set var signature. Fixed tests.

ericsciple:


powershell/Tests/L0/Get-Endpoint.WritesErrorIfRequired.ps1
```diff
@@ -14,7 +14,7 @@ Invoke-VstsTaskScript -ScriptBlock {
         $env:ENDPOINT_AUTH_SOMENAME = $variableSet.Auth
 
         # Act.
-        $actual = Get-VstsEndpoint -Name 'SomeName' -Require 2>&1
+        $actual = Get-VstsEndpoint -Name 'SomeName' -Require -ErrorAction Continue 2>&1
```
ericsciple: I changed the test infrastructure for the task-lib repo in my last merge. There's some slight differences in the infra from the tasks repo, due to the tests in the tasks repo now leveraging Invoke-VstsTaskScript to run the tests. After I changed the infrastructure, a couple tests were failing, so i'm fixing those as part of this PR.

powershell/Tests/L0/Set-TaskVariable.SetsValue.ps1
```diff
@@ -5,19 +5,20 @@ param()
...
-        @{ Path = 'Env:agent.jobstatus' ; Name = 'agent.jobstatus' ; Expected = 'Some job status' }
+        @{ Path = 'Env:Some_name' ; Name = 'Some.name' ; Expected = 'Some value' ; IsSecret = $false }
+        @{ Path = 'Env:Some_name' ; Name = 'Some.name' ; Expected = 'Some value' ; IsSecret = $true }
+        @{ Path = 'Env:agent.jobstatus' ; Name = 'agent.jobstatus' ; Expected = 'Some job status' ; IsSecret = $false }
     )
     foreach ($variableSet in $variableSets) {
-        Unregister-Mock Write-Host
-        Register-Mock Write-Host
+        Unregister-Mock Write-SetVariable
+        Register-Mock Write-SetVariable
```
ericsciple: I found out mocking Write-Host makes things difficult to read the test trace output. Switching over to mocking Write-SetVariable since it makes the assertion easier and doesn't mess up the trace output.

Also, I added IsSecret to the test cases. This was another reason to move away from asserting the value passed to Write-Host. The property bag within the logging command now contains two values (name and issecret). The problem with trying to assert the correct value was passed to write-host is that the property bag internally is passed to a private format method as a hashtable. Ordered hastable wasn't introduced until PowerShell 4, so the order that the properties are written is unpredictable.

powershell/Tests/L0/Write-LoggingCommand.WritesCommand.ps1
```diff
@@ -62,53 +85,48 @@ Invoke-VstsTaskScript -ScriptBlock {
...
             if (!$variableSet.AlternateExpected -or $variableSet.AlternateExpected -ne $actual) {
                 Assert-AreEqual $variableSet.Expected $actual
             }
 
-            # Act/Assert - Verify without using as output switch.
+            # Verify without using as output switch.
+            # Arrange.
+            # Override Write-Host for as short a time as possible. Otherwise the test output becomes confusing.
+            $script:hostMessages = @( )
+            Register-Mock Write-Host { $OFS = " " ; $script:hostMessages += "$args" }
```
ericsciple: I can't avoid mocking write-host in this function since this test is actually asserting the behavior of the private helper function that either writes the logging command to the host or to the output pipeline.

I changed this test so that Write-Host is overridden for as short a time as possible. Overriding it for the full test made the test difficult to debug.

I also added a helper assert method to help out with the unpredictable-order-of-the-logging-command-properties issue and other difficulties asserting write-host was called with the expected value.

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -114,7 +114,8 @@ function Set-TaskVariable {
     param(
         [Parameter(Mandatory = $true)]
         [string]$Name,
-        [string]$Value)
+        [string]$Value,
+        [switch]$Secret)
```
ericsciple: I updated the signature here. The logging command was updated already in the previous merge.

The write-addbuildtag logging command was also added in the previous merge.

@TingluoHuang I tested them E2E and it seems to be correct. In case you want to glance them over:
https://github.com/Microsoft/vsts-task-lib/blob/master/powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
ctrl+f: `write-addbuildtag`
ctrl+f: `write-setvariable`

TingluoHuang:
Looks good.

PR #44, opened 2016-02-21
Users/ersciple/m95locstrings

ericsciple:
Is the fix already in master?  If not, link the bug tracking that work.
No. I've added a linked work item to the bug.

What testing was done?
Tested E2E from an agent on a Linux box connected to a German TFS SKU.

Does this fix change the AT/DT contract?
No.

Does this fix include any database servicing changes?
No. But we might need a new servicing change to get the new tasks into TFS (if one hasn't already been created for the patch milestone). I'll open a bug and follow-up with CAT team.

Were tests added to cover this to make sure it doesn't regress?  If not, link a bug or user story tracking that work.
Yes.

### DIFF
--- a/node/lib/task.ts
+++ b/node/lib/task.ts
@@ -153,14 +153,24 @@ function loadLocStrings(resourceFile: string, culture: string): { [key: string]:
     if (exist(resourceFile)) {
         debug('load strings from: ' + resourceFile);
         var resourceJson = require(resourceFile);
-
         if (resourceJson && resourceJson.hasOwnProperty('messages')) {
             var locResourceJson = null;
             // load up resource resjson for different culture
-            var localizedResourceFile = path.join(path.dirname(resourceFile), 'strings', 'resources.resjson', culture, 'resources.resjson');
-            if (exist(localizedResourceFile)) {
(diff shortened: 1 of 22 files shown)


scottdallamura:
looks good

TingluoHuang:
Looks good.

node/lib/task.ts
```diff
@@ -153,14 +153,24 @@ function loadLocStrings(resourceFile: string, culture: string): { [key: string]:
     if (exist(resourceFile)) {
         debug('load strings from: ' + resourceFile);
         var resourceJson = require(resourceFile);
```
ericsciple: The changes to this file fix the case-sensitivity issue on xplat build agents.

node/package.json
```diff
@@ -38,6 +38,6 @@
     "merge2": "^0.3.6",
     "mocha": "^2.2.5",
     "semver": "^5.1.0",
-    "typescript": "next"
+    "typescript": "1.8.0-dev.20160104"
```
ericsciple: This is a developer tools dependency only. This pins the CI server to a known tsc version that won't break our build.

PR #47, opened 2016-03-03
exit() should stop execution on success

madhurig:


bryanmacfarlane:
We stopped doing this because we were losing log lines by exit'ing the process.  It's deprecated and tasks should use setResult.  I have to back and refresh my memory to check why again ... maybe it's not an issue anymore since we changed the process execution code

madhurig:
After talking to Bryan offline, we decided not to make this change since it causes issues with logger.

PR #52, opened 2016-03-10
fixing tests.

AshwiniChalla:


AshwiniChalla:
eol character is \n for mac OS and I was running tests there before. 
Using _buildOutput function instead of \n to fix it.

prawalagarwal:
Approved.

PR #53, opened 2016-03-11
Support Merge Test Results for Xplat Agent

nigurr:
Presently Xplat agent doesn't support merging of test results.
With this change, Xplat agent will publish merged test results and all the test files will be published in single archive.


acesiddhu:
looks fine

PR #57, opened 2016-03-25
1) Global error pref 2) Secure inputs 3) Mask endpoint auth

ericsciple:


powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -1,3 +1,5 @@
+$script:secureInputs = @{ }
```
ericsciple: This is a new private module-scoped variable that stores the secured input values.

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -15,14 +17,14 @@ function Get-Endpoint {
         [string]$Name,
```
ericsciple: Get-Endpoint and Get-Input has been switched over to store the values securely and wipe the environment variable. Get-TaskVariable has not been switched over and should remain as is.

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -158,60 +160,121 @@ function Get-EnvVariable {
...
+        [object]$Default,
+        [switch]$AsBool,
+        [switch]$AsInt)
+
+    # Attempt to get the secure variable.
+    $value = $null
+    if ($psCredential = $script:secureInputs[$Path]) {
+        $value = $psCredential.GetNetworkCredential().Password
+    } elseif (!$script:nonInteractive) {
+        # The value wasn't found. Prompt for the value if running in interactive dev mode.
```
ericsciple: Interactive dev mode will still actually work nearly the same. With the small caveat that if the developer want to clear one of the values between interactive calls to "Invoke-TaskScript", then they need to un-import the VstsTaskSdk module and re-import it in order to clear the module-scoped dictionary.

For interactive dev mode, if the developer updates a value, then it will just overwrite the existing value in the internal module-scoped dictionary.

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -158,60 +160,121 @@ function Get-EnvVariable {
...
+            $script:secureInputs[$Path] = New-Object System.Management.Automation.PSCredential(
+                $Path,
+                (ConvertTo-SecureString -String $value -AsPlainText -Force))
+        }
+    }
+
+    Get-FinalValue -Value $value @PSBoundParameters
+}
+
+function Get-FinalValue {
```
ericsciple: This new (private) function contains the common code used by both the existing Get-EnvVariable function (called by Get-TaskVariable) and the new Get-SecureInput function (called by Get-Endpoint/Get-Input).

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -158,60 +160,121 @@ function Get-EnvVariable {
...
+        [Parameter(ParameterSetName = 'Require')]
+        [switch]$Require,
+        [Parameter(ParameterSetName = 'Default')]
+        [object]$Default,
+        [switch]$AsBool,
+        [switch]$AsInt)
+
+    $result = $Value
+    if ($result) {
+        if ($Path -like 'Env:ENDPOINT_AUTH_*') {
```
ericsciple: Additional masking for endpoint auth.

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -158,60 +160,121 @@ function Get-EnvVariable {
...
 
-        return $value
+        return $result
     }
 
-    return $value
+    return $result
+}
+
+function Initialize-SecureInputs {
```
ericsciple: Invoke-TaskScript calls this function.

powershell/VstsTaskSdk/ToolFunctions.ps1
```diff
@@ -61,7 +61,9 @@ function Invoke-TaskScript {
         [scriptblock]$ScriptBlock)
 
     try {
+        $global:ErrorActionPreference = 'Stop'
```
ericsciple: Previous the error action preference was set within Merge-Pipelines. I just moved it out for readability. Also it is now set at the global level. This is related to a problem I ran into while testing the Azure PowerShell task changes. The module-scope inherits from the global scope, so errors from within module-scopes weren't getting treated as terminating errors.

I'm still trying to figure out why I didn't catch this earlier. I remember specifically testing this at one point. I need to add a unit test.

powershell/VstsTaskSdk/ToolFunctions.ps1
```diff
@@ -61,7 +61,9 @@ function Invoke-TaskScript {
...
         Write-Verbose "$($MyInvocation.MyCommand.Module.Name) $($MyInvocation.MyCommand.Module.Version)" 4>&1 | Out-Default
+        Initialize-SecureInputs 4>&1 | Out-Default
```
ericsciple: Out-Default isn't fully hijacked yet at this point. I added a very small amount of verbose tracing within Initialize-SecureInputs so it needs to be explicitly piped to the lib's version of Out-Default.

powershell/VstsTaskSdk/ToolFunctions.ps1
```diff
@@ -144,14 +146,14 @@ function Merge-Pipelines {
     param(
         [Parameter(Mandatory = $true)]
         [scriptblock]$ScriptBlock,
-        [switch]$DoNotSetErrorActionPreference)
+        [switch]$NoClosure)
```
ericsciple: I renamed this switch for clarity and moved setting error preference out of this function.

powershell/VstsTaskSdk/VstsTaskSdk.psm1
```diff
@@ -80,5 +80,21 @@ namespace VstsTaskSdk
...
 
+$scriptText = @"
 # Load the SDK resource strings.
-Merge-Pipelines -ScriptBlock ([scriptblock]::Create("Import-LocStrings `"$PSScriptRoot\lib.json`"")) -DoNotSetErrorActionPreference
+Import-LocStrings "$PSScriptRoot\lib.json"
+
+# Load the module that contains ConvertTo-SecureString.
+if (!(Get-Module -Name Microsoft.PowerShell.Security)) {
+    Write-Verbose "Importing the module 'Microsoft.PowerShell.Security'."
+    Import-Module -Name Microsoft.PowerShell.Security 2>&1 |
```
ericsciple: This prevents taking a hit from module-auto loading. A unit test caught this :)

We turn off module auto-loading in the unit tests.

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -158,60 +160,121 @@ function Get-EnvVariable {
...
+    # Store endpoints/inputs in a secure fashion.
+    foreach ($variable in (Get-ChildItem -Path Env:ENDPOINT_*, Env:INPUT_*)) {
+        $path = "Env:$($variable.Name)"
+        if ($variable.Value) {
+            $script:secureInputs[$path] = New-Object System.Management.Automation.PSCredential(
+                $path,
+                (ConvertTo-SecureString -String $variable.Value -AsPlainText -Force))
```
ericsciple: This leverages PowerShell's ConvertTo-SecureString function to store the string securely.

The lookup key for the hashtable is the drive-qualified path of the ENV variable (e.g. `Env:INPUT_FOO`). Using this as the lookup key required the least modifications to the existing code and allows more code re-use between Get-TaskVariable and Get-Endpoint/Get-Input.

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -158,60 +160,121 @@ function Get-EnvVariable {
...
+        return $result
     }
 
-    return $value
+    return $result
+}
+
+function Initialize-SecureInputs {
+    # Store endpoints/inputs in a secure fashion.
+    foreach ($variable in (Get-ChildItem -Path Env:ENDPOINT_*, Env:INPUT_*)) {
```
TingluoHuang: is this means we will put all inputs into security string, even the input is not a secret?

ericsciple: That's right. There's no way of knowing whether the input has a secret or not. This also solves the problem that when a task invokes an external process (toolrunner), the external process should not have access to any of the INPUT/ENDPOINT variables.

ericsciple: Note, the value is still traced to debug when the task author calls Get-Input

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -67,7 +69,7 @@ function Get-Input {
         [switch]$AsInt)
 
     $PSBoundParameters['Name'] = (Get-LocString -Key PSLIB_Input0 -ArgumentList $Name)
-    Get-EnvVariable @PSBoundParameters -Path "Env:INPUT_$($Name.Replace(' ', '_').ToUpperInvariant())"
+    Get-SecureInput @PSBoundParameters -Path "Env:INPUT_$($Name.Replace(' ', '_').ToUpperInvariant())"
```
TingluoHuang: you had a func: function Format-VariableName in the bottom. :D

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -67,7 +69,7 @@ function Get-Input {
         [switch]$AsInt)
```
TingluoHuang: Where did you use $Require/AsBool/AsInt in Get-Input?

ericsciple: It's passed when calling Get-SecureInput by splatting the PSBoundParameters automatic variable (using the `@` splat operator).

powershell/VstsTaskSdk/InputFunctions.ps1
ericsciple: The func at the bottom is common code shared between Get-TaskVariable and Set-TaskVariable. For some reason the other functions (Get-Input/Get-Endpoint) have different formatting rules.

powershell/VstsTaskSdk/InputFunctions.ps1
```diff
@@ -158,60 +160,121 @@ function Get-EnvVariable {
...
+        [switch]$Require,
+        [Parameter(ParameterSetName = 'Default')]
+        [object]$Default,
+        [switch]$AsBool,
+        [switch]$AsInt)
+
+    # Attempt to get the secure variable.
+    $value = $null
+    if ($psCredential = $script:secureInputs[$Path]) {
+        $value = $psCredential.GetNetworkCredential().Password
```
TingluoHuang: I can't find any related documentation, will GetNetworkCredential() always work? even on a not domain joined machine?

ericsciple: Good question. I'll just confirmed it will work on my non-domain joined machine at home. FYI - I peeked into the .Net Core source code of GetNetworkCredential when I was adding this... It is surprisingly simple. It's basically just a class with three string properties (domain, username, pw). I had expected to find more magic :)

TingluoHuang:
Looks good. :D

PR #70, opened 2016-04-20
Add agent dependency info.

ericsciple:


ericsciple:
The handler isn't in the Windows agent in the releases/m89 branch, so it must have missed Update 1 for the Windows agent. I didn't have success running the genmd.js script, so I wasn't able to generate the md.

PR #76, opened 2016-05-04
Update getDelimitedInput to remove empty values.

ericsciple:


node/test/tasklib.ts
```diff
@@ -564,6 +596,32 @@ describe('Test vsts-task-lib', function() {
 
             done();
         })
+
+        // filePathSupplied tests
```
ericsciple: I just moved these two tests. I was trying to figure out what was covered already.

The tests were largely grouped already by the function being tested. Only these two were out of place.

bryanmacfarlane: OK - I have a note to move into separate files/suites as well

bryanmacfarlane:
Need to bump version in package.json

bryanmacfarlane:
Thanks

bryanmacfarlane:
Published to npm.

PR #77, opened 2016-05-05
Added release notes.

ericsciple:


node/docs/genmd.js
```diff
@@ -94,14 +94,16 @@ var ds = require('./structure.json');
 writeLine('# VSTS-TASK-LIB TYPESCRIPT API');
 writeLine();
 writeLine('## Dependencies');
-writeLine('A [cross platform agent](https://github.com/Microsoft/vso-agent) OR a TFS 2015 Update 2 Windows agent (or higher) is required to run a Node task end-to-end. However, an agent is not required for interactively testing the task.');
+writeLine('A [cross platform agent](https://github.com/Microsoft/vsts-agent) OR a TFS 2015 Update 2 Windows agent (or higher) is required to run a Node task end-to-end. However, an agent is not required for interactively testing the task.');
```
ericsciple: On second thought, reverting this.

node/docs/genmd.js
```diff
@@ -102,6 +102,8 @@ writeLine();
 writeLine('The build generates a vsts-task-lib.d.ts file for use when compiling tasks');
 writeLine('In the example below, it is in a folder named definitions above the tasks lib');
 writeLine();
+writeLine('## [Release notes](releases.md)');
```
ericsciple: I'm unable to validate this change. I get errors on Windows when attempting to run docs/genmd.js directly or when attempting to run docs.sh.
```
Error: C:/vsts-task-lib/node/typings/main/ambient/node/index.d.ts(84)
 Type expected.
Error: C:/vsts-task-lib/node/typings/main/ambient/node/index.d.ts(187)
 Type expected.
Error: C:/vsts-task-lib/node/typings/main/ambient/node/index.d.ts(188)
 Type expected.
[omitted about 20 more]
```

bryanmacfarlane:
don't revert.  gendocs is broken

node/docs/genmd.js
bryanmacfarlane: gendocs calls into typedoc which hasn't caught up to ts compiler we need.  I've been hand editing docs since.  You're not changing signature - just impl so no need for docs update unless you want to directly update the comments

ericsciple:
I see. In that case I'll also update the MD by hand.

I had originally updated the link from the vso-agent to the vsts-agent repo... although I reverted that. I think the vso-agent repo has links to the new one anyway.

node/docs/genmd.js
ericsciple: I see. I forgot about the comments on the function itself. Mainly my intended goal is to start a release notes MD. That way task authors can better understand the implications of picking up a new version of the lib.

bryanmacfarlane:
Let's just start as a separate md which we manually writeup

ericsciple:
Yep, I did. I only made two edits to vsts-task-lib.md:
1) Added a link to the separate release notes markdown.
2) Hand edited to include the "Dependencies" section. I had previously added this to genmd.js, but it has never been regenerated since.

Might be good to touch base tomorrow to make sure what I'm envisioning is what you are envisioning.

PR #80, opened 2016-05-10
Updated set result failed to create an error issue.

ericsciple:


bryanmacfarlane:
I released 0.7.3 to npm.  We can update tasks today and do some E2E testing on bh8

PR #81, opened 2016-05-12
Fix Windows which

ryuyu:
There was a bug here in which the isFile checks were not actually being called.
This led to a relatively obscure bug where if the command searched for was not found in the path, and shared a name with a folder in the current directory, that folder was returned as the value of which.
This PR fixes this issue.

That being said, the fully correct fix for this issue would be to upgrade shelljs and remove this windows specific code as shelljs has fixed the indicated bug. I have not done this here as I am unsure what the larger repercussions of that upgrade may be.

@bryanmacfarlane

### DIFF
--- a/node/lib/task.ts
+++ b/node/lib/task.ts
@@ -646,17 +646,17 @@ export function which(tool: string, check?: boolean): string {
 
                     var baseAttempt = attempt;
                     attempt = baseAttempt + '.exe';
-                    if (exist(attempt) && stats(attempt).isFile) {
+                    if (exist(attempt) && stats(attempt).isFile()) {
                         toolPath = attempt;
                         return;
                     }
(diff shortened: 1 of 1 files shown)


bryanmacfarlane:
Ting?

bryanmacfarlane:
I thought new shelljs would allow us to get rid of this code ... Ting can look

ryuyu:
That is correct. As noted above, upgrading shelljs to (I believe) 0.7.0+ would fix this inside shelljs itself, and we could just drop this block altogether in favour of the shelljs implementation. I was afraid to that as i was unsure of the other ramifications as shelljs is a pretty integral part of the way this package works.

ericsciple:
rewrote which: https://github.com/Microsoft/vsts-task-lib/pull/231

PR #83, opened 2016-05-30
Handle Mixed File Path separators in getPathInput

nigurr:
Converts file paths to Posix path which are usually supported in windows. This will convert mixed file path seperators to posix type.


bryanmacfarlane:
What problem are you solving?  The file paths as they are now work fine on windows when run by node in node paths.

nigurr:
Assume there's a Maven build definition, where user has given build file location as "TestProj\pom.xml", This works fine in Windows Agent, however this build will break in coreCLR agent if user is running on *nix machines. *nix can't handle "\" slash. this will be common behavior in case of VsTest task (future)

bryanmacfarlane:
I want to think about this a bit more.  We've thrown stuff into the lib before only to pull back and change.  It's not clear whether we should fix on server or agent (and not in lib).  Back slash also means something special on nix (escape) but have to go through scenarios to see if it matters on a path

ericsciple:
Setting aside the "back slash means something on nix" issue...

This statement concerns me: "usually supported in Windows".

Forward slashes do not work cmd.exe. If you do any sort of optimistic conversion, wouldn't you convert to the separator of the current platform (`/` on nix, `\` on Windows)?

nigurr:
@ericsciple `/` do works on cmd and powershell in windows. 

I would be happy if we can convert to os dependent separator.

ericsciple:
> `/` do works on cmd and powershell in windows.

Two examples that do not work in cmd.exe:
1)
```
D:\vsts-agent-2.102.0>./run.cmd
'.' is not recognized as an internal or external command,
operable program or batch file.
```

2)
```
C:\>dir /users
Invalid switch - "users".
```

nigurr:
Interesting. I guess few commands messed up. I will see if we can handle OS specific way

C:\Users\nigurr>cd /users

C:\Users>

nigurr:
@ericsciple changed the functionality. please verfiy

ericsciple:
Can we hold this for a month? I'm following a similar issue in a different project to see how they handle the same problem

nigurr:
@ericsciple, probably we can adopt the implementation if we find something better? I feel we can adopt this as it's covering all the cases.

ericsciple:
The problem is, it doesn't cover all the cases. `\` is a valid file name character in OSX/Linux.

bryanmacfarlane:
I don't want to take the approach of adopting something better if it comes along.  This is in the base, base lib of all tasks in something as core as path handling.  I want Eric to look at other examples for how to do this reliably.

bryanmacfarlane:
1. If it's a file path input, this is an agent problem, not a task lib problem.
2. having one def that works across all OSes isn't desirable and we even stated it as a non goal (initially we discussed) because it makes reporting hard.  What you want is BuildDef-Win, Build-Def-OSX etc... so you can see that the build/tests are broken on OSX.  All our reporting is definition level.  Practically, we don't want to do this.  To make that easier, we will add process inheritance.
3.  If you really want a path agnostic path then use /.  Node handles that the same across all platforms.

Finally, if this is future vstest, then I don't want to take for Dev15.

bryanmacfarlane:
Closing per my points above.

Note we did take this change.  Feel free to follow up with Madhuri on details of all our discussions: https://github.com/Microsoft/vsts-task-lib/pull/101

PR #85, opened 2016-06-01
Update File Exist Function

nigurr:
Upgrade the exist function since node deprecated thae existsSync

### DIFF
--- a/node/lib/task.ts
+++ b/node/lib/task.ts
@@ -511,7 +511,17 @@ export function stats(path: string): FsStats {
  * @returns   boolean 
  */
 export function exist(path: string): boolean {
-    return path && fs.statSync(path) != null;
+    var exist = false;
+    try {
+        exist = path && fs.statSync(path) != null;
+    } catch (err) {
+        if (err && err.code === 'ENOENT') {
+            exist = false;
+        } else {
+            throw err;
+        }
+    }
+    return exist;
 }
 
 /**
--- a/node/lib/task.ts
+++ b/node/lib/task.ts
@@ -511,7 +511,11 @@ export function stats(path: string): FsStats {
  * @returns   boolean 
  */
 export function exist(path: string): boolean {
-    return path && fs.existsSync(path);
+    try {
+        return path && fs.statSync(path) != null;
+    } catch (err) {
+        return false;
+    }
 }
 
 /**
--- a/node/lib/task.ts
+++ b/node/lib/task.ts
@@ -511,11 +511,7 @@ export function stats(path: string): FsStats {
  * @returns   boolean 
  */
 export function exist(path: string): boolean {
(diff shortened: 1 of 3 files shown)


node/lib/task.ts
```diff
@@ -511,7 +511,11 @@ export function stats(path: string): FsStats {
  * @returns   boolean 
  */
 export function exist(path: string): boolean {
-    return path && fs.existsSync(path);
+    try {
+        return path && fs.statSync(path) != null;
+    } catch (err) {
+        return false;
```
bryanmacfarlane: We can't do this.  You are conclusively saying it doesn't exist if it throws.  For example, if you don't have permissions to the file, then you will eat that and return conclusively that it doesn't exist.  That's not true.

Instead the task should just use exist inline.  That task should be wrapped in one outer try/catch (see ShellScript task).  If it fails to check if a file exists, it should fail.  And it will fail with a clear message EPERM in my example instead of hiding the issue leading to a harder to diagnose issue downstream when the task proceeds to do work on the premise that it doesn't exist.

nigurr: Seems like this works for "right" for checking whether file exists

>>sudo touch test.txt
>>sudo chmod 000 test.txt
>>node test.js test.txt  //returns true the current user is not root
>> ls /etc/usr/bin
ls: cannot access /etc/usr/bin: No such file or directory
>> node test.js /etc/usr/bin //return true

bryanmacfarlane: Do not swallow exceptions.  You've just masked whatever the exceptional case is and returned a conclusive answer which s not necessarily correct.

bryanmacfarlane: Please identify what specific issue/case you're addressing which led you to want to change this code

nigurr:
Please review the change

node/test/tasklib.ts
```diff
@@ -1565,5 +1566,18 @@ describe('Test vsts-task-lib', function() {
             assert(expectedOutput === output, _mismatch(expectedOutput, output));
             done();
         })
+
+        it('check exist functionality', function (done) {
```
bryanmacfarlane: split into "it checks file exists" and "it checks file does not exist"

node/test/tasklib.ts
```diff
@@ -19,47 +19,47 @@ var _testTemp = path.join(__dirname, '_temp');
 
 var plat = os.platform();
 
-var NullStream = function() {
+var NullStream = function () {
```
bryanmacfarlane: all the formatting changes are OK but it makes focused changes really hard to review :)

bryanmacfarlane:
address merge conflicts (likely from formatting), split test and OK

nigurr:
Addressed Code review comments

PR #86, opened 2016-06-01
Update FindFunctions.ps1

mthalman-msft:
When using the IncludeDirectories switch parameter, it would result in a null ref error in the Get-PathIterator function.  This is due to the use of a variable that doesn't exist: "$i_".  Corrected it by changing it to "$_" instead.


ericsciple:
@mthalman-msft Thanks!

PR #89, opened 2016-06-03
Secret variable.

ericsciple:


ericsciple:
@TingluoHuang feel free to merge and finish off. I only verified it doesn't doesn't break any existing tests. Thankfully they are there - I broken them multiple times trying to get it right :)

I didn't add any new tests, and I didn't test end-to-end. I didn't even test the functionality works, it should be pretty close if it doesn't work though.

ericsciple:
Tests added now.

PR #95, opened 2016-06-15
adding mock hooks & null check in toolrunner

jasholl:
Need to expose the following methods in task so I can mock them in L0 tests in vsts-tasks

process.cwd(): string;
os.type(): string;
fs.readdirSync(path: string): string[];

Also adding a null check in tool runner so if the tool is not kicked off properly and results in no stdout or stderr, node doesn't just die but instead bubbles up a failure to the offending task.

### DIFF
--- a/node/lib/task.ts
+++ b/node/lib/task.ts
@@ -591,6 +591,26 @@ export function exist(path: string): boolean {
 }
 
 /**
+ * Useful for determining the host operating system.
+ * see [os.type](https://nodejs.org/api/os.html#os_os_type)
+ * 
+ * @return      the name of the operating system
+ */
+export function osType(): string {
+    return os.type();
+}
+
+/**
+ * Returns the process's current working directory.
+ * see [process.cwd](https://nodejs.org/api/process.html#process_process_cwd)
+ * 
+ * @return      the path to the current working directory of the process
+ */
+export function cwd() : string {
+    return process.cwd();
+}
+
+/**
  * Checks whether a path exists.
  * If the path does not exist, the task will fail with an error message. Execution will halt.
  * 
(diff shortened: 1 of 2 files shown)


node/lib/task.ts
```diff
@@ -591,6 +591,47 @@ export function exist(path: string): boolean {
 }
 
 /**
+ * Reads a directory's contents.
+ * see [fs.readdirsync](https://nodejs.org/api/fs.html#fs_fs_readdirsync_path_options)
+ * 
+ * @param       path    path of directory to readdirSync
+ * @returns     array of directory contents
+ */
+export function readdirSync(path: string): string[] {
```
bryanmacfarlane: we have been wrapping shelljs with other disk operations.  That would mean we add ls function, not readdirSync.

node/lib/task.ts
```diff
@@ -591,6 +591,47 @@ export function exist(path: string): boolean {
...
+}
+
+/**
+ * Deletes the specified file.
+ * see [fs.unlinkSync](https://nodejs.org/api/fs.html#fs_fs_unlinksync_path)
+ * 
+ * @param        path   path of the file to delete
+ */
+export function unlinkSync(path: string): void {
+    fs.unlinkSync(path);
```
bryanmacfarlane: shelljs rm?  We already have rmrf function.

node/lib/task.ts
```diff
@@ -591,6 +591,47 @@ export function exist(path: string): boolean {
...
+    return os.type();
+}
+
+/**
+ * Returns the process's current working directory.
+ * see [process.cwd](https://nodejs.org/api/process.html#process_process_cwd)
+ * 
+ * @return      the path to the current working directory of the process
+ */
+export function processCwd() : string {
```
bryanmacfarlane: would prefer cwd method

node/lib/toolrunner.ts
```diff
@@ -376,8 +377,8 @@ export class ToolRunner extends events.EventEmitter {
         }
 
         var res:IExecResult = <IExecResult>{ code: r.status, error: r.error };
-        res.stdout = r.stdout.toString();
-        res.stderr = r.stderr.toString();
+        res.stdout = (r.stdout) ? r.stdout.toString() : null;
+        res.stderr = (r.stderr) ? r.stderr.toString() : null;
```
bryanmacfarlane: ok

node/lib/task.ts
jasholl: added ls instead.

node/lib/task.ts
jasholl: didn't see that method before; switched to using rmRF and pulled readdirSync out.

node/lib/task.ts
jasholl: added cwd instead as requested.

PR #98, opened 2016-07-04
Fixed Get-VstsVssCredentials.

ericsciple:


bryanmacfarlane:
Non issue if we package libs with tasks

PR #102, opened 2016-07-19
OM functions.

ericsciple:
Soliciting early feedback. I still need to test against Dev14 and tweak the Extended Client function.

Currently, this would reduce client code to load an HTTP client to simply:
```powershell
$projectHttpClient = New-VstsVssHttpClient -TypeName Microsoft.TeamFoundation.Core.WebApi.ProjectHttpClient
$projectHttpClient.GetProjects().GetAwaiter().GetResult()
```

### DIFF
--- a/powershell/VstsTaskSdk/ServerOMFunctions.ps1
+++ b/powershell/VstsTaskSdk/ServerOMFunctions.ps1
@@ -5,27 +5,130 @@ Gets a credentials object that can be used with the TFS extended client SDK.
 .DESCRIPTION
 The agent job token is used to construct the credentials object. The identity associated with the token depends on the scope selected in the build/release definition (either the project collection build/release service identity, or the project build/release service identity).
 
+Refer to Get-VstsTfsService for a more simple to get a TFS service object.
+
+*** DO NOT USE Agent.ServerOMDirectory *** See TODO_HREF for reliable usage when working with the TFS extended client SDK from a task.
+
+.PARAMETER OMDirectory
+Directory where the extended client object model DLLs are located. If the DLLs for the credential types are not already loaded, an attempt will be made to automatically load the required DLLs from the object model directory.
+
+If not specified, defaults to the directory of the entry script for the task.
+
+*** DO NOT USE Agent.ServerOMDirectory *** See TODO_HREF for reliable usage when working with the TFS extended client SDK from a task.
+
 .EXAMPLE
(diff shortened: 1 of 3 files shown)


powershell/VstsTaskSdk/ServerOMFunctions.ps1
```diff
@@ -48,10 +48,145 @@ $projectHttpClient.GetProjects().Result
...
+        [uri]$Uri = New-Object System.Uri($Uri)
+
+        if (!$VssCredentials) {
+            # Default the credentials.
+            $VssCredentials = Get-VssCredentials
+        }
+
+        # Try to load the type.
+        try {
+            $null = [type]$TypeName
```
bgavrilMS: You know too much PS. I think $TypeName -is [Type] would be somewhat more readable.

powershell/VstsTaskSdk/ServerOMFunctions.ps1
```diff
@@ -48,10 +48,145 @@ $projectHttpClient.GetProjects().Result
...
+            Write-Verbose "Testing file path: '$newtonsoftDll'"
+            if (!(Test-Path -LiteralPath $newtonsoftDll -PathType Leaf)) {
+                Write-Verbose 'Not found. Rethrowing exception.'
+                throw
+            }
+
+            # Add a binding redirect and try again. Parts of the Dev15 preview SDK have a
+            # dependency on the 6.0.0.0 Newtonsoft.Json DLL, while other parts reference
+            # the 8.0.0.0 Newtonsoft.Json DLL.
+            Write-Verbose "Adding assembly resolver."
```
bgavrilMS: I would add the assembly resolver logic unconditionally, just like the VsoWorker does through its vsoworker.exe.config. There's also a redirect for System.Net.Http.Formatting which we haven't crashed on, but it may be worth adding.

bgavrilMS:
LGTM

ericsciple:
@TingluoHuang this is ready for review now. The function signatures and logic is all ready for review.

I need to remove commented-out code and fixup the help comments (synopsis/description/example comments exposed via the powershell help system).

I'm going to need to bump the minor version number as part of this commit, since it is a functional change (defaulted to serveromdirectory before, but that is fragile and we need to discourage folks from doing that).

After this is merged I need to change the publish symbols task to be a reference example. And then I need to add docs in this repo to provide guidance on the safe way to consume the client OM.

PR #103, opened 2016-07-19
Fix issue in generating Command for boolean values

nigurr:
if(val) falsify the boolean parameters. 
We should not send command only for undefined/null types

### DIFF
--- a/node/taskcommand.ts
+++ b/node/taskcommand.ts
@@ -31,7 +31,7 @@ export class TaskCommand {
             for (var key in this.properties) {
                 if (this.properties.hasOwnProperty(key)) {
                     var val = this.properties[key];
-                    if (val) {
+                    if (val != null) {
                         cmdStr += key + '=' + val + ';';
                     }                    
                 }
--- a/node/test/tasklib.ts
+++ b/node/test/tasklib.ts
@@ -803,6 +803,20 @@ describe('Test vsts-task-lib', function () {
             assert.equal(tc.toString(), '##vso[some.cmd]a message');
             done();
         })
+        it('handles boolean properties', function (done) {
+            this.timeout(1000);
+
+            var tc = new tcm.TaskCommand('some.cmd', { foo: false }, 'a message');
+            assert.equal(tc.toString(), '##vso[some.cmd foo=false;]a message');
+            done();
+        })
+        it('handles undefined properties', function (done) {
+            this.timeout(1000);
+
+            var tc = new tcm.TaskCommand('some.cmd', { foo: undefined }, 'a message');
+            assert.equal(tc.toString(), '##vso[some.cmd ]a message');
+            done();
+        })
         it('parses cmd with no properties', function (done) {
             var cmdStr = '##vso[basic.command]messageVal';
 


node/taskcommand.ts
```diff
@@ -31,7 +31,7 @@ export class TaskCommand {
             for (var key in this.properties) {
                 if (this.properties.hasOwnProperty(key)) {
                     var val = this.properties[key];
-                    if (val) {
+                    if (val != null) {
```
ericsciple: Would this line also need to change:
```typescript
public properties: {[key: string]: string};
```

bryanmacfarlane:
What issue is driving this?

jpricketMSFT:
It seems like bad practice to use null in TypeScript or JavaScript. Undefined is the standard and doesn't equal null.

nigurr:
@jpricketMSFT accept that from best practices. but `if(var != null)` does undefined and null check in `typescript`. So i would take the safe check rather than flowing it down.

@bryanmacfarlane, you can check the UT which added. If `you are trying to pass boolean values in properties [foo = false], the if check falsify that property and don't write to the command at all`

nigurr:
We merge publishing test reports by default. If one doesn't want to avoid, the task has to send "mergeReports=false", which is being skipped by this command because of the bug.

node/test/tasklib.ts
```diff
@@ -803,6 +803,20 @@ describe('Test vsts-task-lib', function () {
             assert.equal(tc.toString(), '##vso[some.cmd]a message');
             done();
         })
+        it('handles boolean properties', function (done) {
+            this.timeout(1000);
+
+            var tc = new tcm.TaskCommand('some.cmd', { foo: false }, 'a message');
```
bryanmacfarlane: properties are a string, string dictionary as Eric noted above.  I'm confused why you're passing a boolean.  See TaskCommand.ts

node/taskcommand.ts
bryanmacfarlane: I'm still not getting why this is needed

```
$ node
> var properties={}
> var val = properties['foo'];
> if (val) { console.log('yep'); } else { console.log('nope'); }
nope
```

So, it seems to do it's job of only running if the key existed.

node/test/tasklib.ts
nigurr: Agree with that. Since the `string` this is at compile time and task can actually send command like mentioned in UT at run time. This check skips that property and doesn't send, which shouldn't be the case.

node/taskcommand.ts
nigurr: > var properties={}
> properties['foo']=false
> properties
{ foo: false }
> var val = properties['foo']
> if(val) { "this is still valid value, not a bool check" } else {"missing value"}
missing value

bryanmacfarlane: They did something wrong, it should be skipped.  Writing code to enable passing the wrong data type encourages doing the wrong thing.  The compiler will encourage the right thing.  If they bypass that and do the wrong thing, then it not getting set will further encourage the right thing.  They will see it's not getting set, look at the docs/types and fix it.

nigurr: makes sense. thanks :)
i will discard this pr

PR #104, opened 2016-07-19
Change Tool discovery order

nigurr:
Current order doesn't work with Ant tool. Ant.cmd don't build as expected to be. Ant.bat does that

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -730,12 +730,12 @@ export function which(tool: string, check?: boolean): string {
                         toolPath = attempt;
                         return;
                     }
-                    attempt = baseAttempt + '.cmd';
+                    attempt = baseAttempt + '.bat';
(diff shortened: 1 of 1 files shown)


jpricketMSFT:
Looks good to me. Ting had mentioned that we should be able get rid of our "which" logic for Windows. Is there a plan to do that work?

bryanmacfarlane:
Yes.  We need to pick up new shelljs.  We should do that instead

But its core so would prefer 104

Get Outlook for iOS<https://aka.ms/o0ukef>




On Tue, Jul 19, 2016 at 9:37 AM -0400, "Jason Prickett MSFT" <notifications@github.com<mailto:notifications@github.com>> wrote:


Looks good to me. Ting had mentioned that we should be able get rid of our "which" logic for Windows. Is there a plan to do that work?

-

PR #108, opened 2016-07-26
Variable names.

ericsciple:
- [x] Bump versions
- [x] Make existing tests pass
- [x] Test E2E
- [x] Add new tests
- [x] Update docs

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -17,7 +17,18 @@ export enum TaskResult {
     Failed = 1
 }
 
-var vault: vm.Vault;
+/**
+ * Hash table of known variable info. The formatted env var name is the lookup key.
+ *
+ * The purpose of this hash table is to keep track of known variables. The hash table
+ * needs to be maintained for multiple reasons:
+ *  1) to distinguish between env vars and job vars
+ *  2) to distinguish between secret vars and public
+ *  3) to know the real variable name and not just the formatted env var name.
+ */
+let knownVariableMap: { [key: string]: KnownVariableInfo; } = { };
+
+let vault: vm.Vault;
 
 //-----------------------------------------------------
 // String convenience
@@ -243,19 +254,33 @@ export function loc(key: string, ...param: any[]): string {
 // Input Helpers
 //-----------------------------------------------------
 /**
- * Gets a variables value which is defined on the build definition or set at runtime.
+ * Gets a variable value that is defined on the build/release definition or set at runtime.
  * 
  * @param     name     name of the variable to get
  * @returns   string
  */
 export function getVariable(name: string): string {
-    var varval;
-    if (vault) {
-        varval = vault.retrieveSecret('SECRET_' + name.replace(/\./g, '_').toUpperCase());
+    let varval: string;
+
+    // get the metadata
+    let info: KnownVariableInfo;
+    let key: string = getVariableKey(name);
+    if (knownVariableMap.hasOwnProperty(key)) {
+        info = knownVariableMap[key];
     }
 
-    if (!varval) {
-        varval = process.env[name.replace(/\./g, '_').toUpperCase()];
+    if (info && info.secret) {
+        // get the secret value
+        varval = vault.retrieveSecret('SECRET_' + key);
+    }
+    else {
+        // get the public value
+        varval = process.env[key];
+
+        // fallback for pre 2.104.1 agent
+        if (!varval && name.toUpperCase() == 'AGENT.JOBSTATUS') {
+            varval = process.env['agent.jobstatus'];
+        }
     }
 
     debug(name + '=' + varval);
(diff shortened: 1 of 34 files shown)


ericsciple:
@madhurig fyi

node/task.ts
```diff
@@ -17,6 +17,11 @@ export enum TaskResult {
     Failed = 1
 }
 
+/** Hash table of variable info. Only the 'name' and 'secret' properties are populated
+ *  on each VariableInfo.
+ */
+var variableInfo: { [key: string]: VariableInfo; } = { };
```
bryanmacfarlane: nit - it's not a variableInfo, it's a map of variableInfos.  Also, we should switch to using typescript let.  So?

```
let variableInfoMap: ...
```

node/task.ts
```diff
@@ -243,41 +248,110 @@ export function loc(key: string, ...param: any[]): string {
...
 export function getVariable(name: string): string {
     var varval;
-    if (vault) {
-        varval = vault.retrieveSecret('SECRET_' + name.replace(/\./g, '_').toUpperCase());
+
+    // get the metadata
+    var info: VariableInfo;
+    var key: string = getVariableKey(name);
+    if (variableInfo.hasOwnProperty(key)) {
+        info = variableInfo[key];
```
bryanmacfarlane: Why have the hasOwnProperty check?  Either way, if it's not in map, info will be undefined

```
> var foo = {};
> var variableInfoMap = {};
> var info = variableInfoMap['foo'];
> info
undefined
```

node/task.ts
```diff
@@ -243,41 +248,110 @@ export function loc(key: string, ...param: any[]): string {
...
     }
 
-    if (!varval) {
-        varval = process.env[name.replace(/\./g, '_').toUpperCase()];
+    if (info && info.secret) {
+        // get the secret value
+        varval = vault.retrieveSecret('SECRET_' + key);
+    } else {
```
bryanmacfarlane: nit
```
}
else {
```

node/task.ts
```diff
@@ -243,41 +248,110 @@ export function loc(key: string, ...param: any[]): string {
...
     }
 
     debug(name + '=' + varval);
     return varval;
 }
 
 /**
- * Sets a variables which will be available to subsequent tasks as well.
+ * Gets all variables that are defined on the build definition or set at runtime.
+ * Requires a 2.104.1 agent or higher.
```
bryanmacfarlane: I still wonder if we shouldn't have a function
```
demand('2.104.1');
```
Which will error if it's not true.  If a task author uses and they don't demand the agent, this will protect the consumer and let them know they need to up there agent instead of the task just not working leading to a long investigation.

node/task.ts
```diff
@@ -243,41 +248,110 @@ export function loc(key: string, ...param: any[]): string {
...
+ * 
+ * @returns VariableInfo[]
+ */
+export function getVariables(): VariableInfo[] {
+    return Object.keys(variableInfo)
```
bryanmacfarlane: demand('2.104.1');

node/task.ts
```diff
@@ -243,41 +248,110 @@ export function loc(key: string, ...param: any[]): string {
...
 }
 
 /**
- * Sets a variables which will be available to subsequent tasks as well.
+ * Gets all variables that are defined on the build definition or set at runtime.
+ * Requires a 2.104.1 agent or higher.
+ * 
+ * @returns VariableInfo[]
+ */
+export function getVariables(): VariableInfo[] {
```
bryanmacfarlane: demand('2.104.1');

node/task.ts
```diff
@@ -243,41 +248,110 @@ export function loc(key: string, ...param: any[]): string {
...
+    return Object.keys(variableInfo)
+        .filter(function (key: string) {
```
bryanmacfarlane: .filter((key: string) => {

node/task.ts
```diff
@@ -243,41 +248,110 @@ export function loc(key: string, ...param: any[]): string {
...
+    var varValue = val || '';
+    debug('set ' + name + '=' + (secret && varValue ? '********' : varValue));
+    if (secret) {
+        vault.storeSecret('SECRET_' + key, varValue);
+    } else {
+        process.env[key] = varValue;
+    }
+
+    // store the metadata
+    variableInfo[key] = new VariableInfo(name, undefined, secret);
```
bryanmacfarlane: Why do we set the value as undefined?

node/task.ts
```diff
@@ -243,41 +248,110 @@ export function loc(key: string, ...param: any[]): string {
...
+
+/** Snapshot of a variable at the time when getVariables was called. */
+export class VariableInfo {
+    constructor(name: string, value: string, secret: boolean) {
+        this.name = name;
+        this.value = value;
+        this.secret = secret;
+    }
+
+    /** Variable name */
```
bryanmacfarlane: nit - comments seems kind of obvious :)  Not sure if needed

node/task.ts
```diff
@@ -1107,6 +1181,20 @@ export function _loadData(): void {
...
+            // Record the secret variable metadata. This is required by getVariable to
+            // retrieve the value. In a 2.104.1 agent or higher, this metadata will be overwritten
+            // when the VSTS_SECRET_VARIABLES env var is processed.
+            if (envvar.startsWith('SECRET_')) {
+                var variableName: string = envvar.substring('SECRET_'.length);
+                if (variableName) {
+                    // This is technically not the variable name (has underscores instead of dots),
+                    // but it's good enough to make getVariable work in a pre-2.104.1 agent where
+                    // the VSTS_SECRET_VARIABLES env var is not define.
+                    variableInfo[getVariableKey(variableName)] = new VariableInfo(variableName, undefined, true);
```
bryanmacfarlane: same questions about setting value to undefined

node/task.ts
```diff
@@ -1116,6 +1204,20 @@ export function _loadData(): void {
         }
     }
     debug('loaded ' + loaded);
+
+    // store public variable metadata
+    JSON.parse(process.env['VSTS_PUBLIC_VARIABLES'] || '[]')
```
bryanmacfarlane: I typically protect JSON.parse because it can throw but I guess this isn't user data (config file etc...).  We should always write valid JSON if it's set ...

node/task.ts
ericsciple: The function requires a 2.104.1 agent to be full featured, but may still be useful on a pre-2.104.1 agent. Let's chat about it, I also don't want to overcomplicate it.

node/task.ts
ericsciple: This is confusion because of the way I implemented it. I think I see a better alternative now.

The point here is to keep track of a list of known variables. We need to maintain this list for several reasons (to distinguish between env vars and job vars; to distinguish between secret vars and public; to know the real variable name and not just the formatted env var name).

All that I really want to record here is variable name, and whether it's a secret. I don't want to run into multi-mastering problems by storing the value (env var stores the correct value for public vars, vault stores the correct value for secret vars).

I think it may improve readability to use a separate class "KnownVariable" to keep track of this info.

node/task.ts
ericsciple: I thought hasOwnProperty precaution is required due other folks extending base classes via prototype or something.

node/task.ts
ericsciple: I'll wrap in try catch for the interactive dev scenario - to gives the developer additional context what is blowing up. No need to loc the message.

PR #109, opened 2016-08-04
wrap fs.writeFileSync

madhurig:


node/task.ts
```diff
@@ -588,6 +588,10 @@ export function exist(path: string): boolean {
     return exist;
 }
 
+export function writeFileSync(file: string, data: string, options? : any) {
```
bryanmacfarlane: Take Sync out of the name.  Even though that's the node pattern, the lib follows the shelljs/shell pattern.  mv(), cp() etc...

We should define an interface for options instead of any.
I also think data is any (takes a string | buffer)

here's the node.d.ts interface
```
export function writeFileSync(filename: string, data: any, options?: { encoding?: string; mode?: number; flag?: string; }): void;
```

bryanmacfarlane: technically we can do data: string | Buffer
Buffer is defined in node and we already leak node types (FsStats)

bryanmacfarlane:
OK.

bryanmacfarlane:
Can you add some unit tests?

madhurig:
Added a couple of tests, not verifying all cases since we just call into fs.writeFileSync()

PR #110, opened 2016-08-04
move tests into appropriate suites

madhurig:


bryanmacfarlane:
BTW, I've wanted to split the suites into separate files ...

PR #111, opened 2016-08-11
Fix find.

ericsciple:
- [x] I need to do more manual testing with `rm -f` to make sure it's going to suffice for what I need from the Delete Files task. It needs to handle deleting symlink folders/files (i.e. delete the symlink only, not the actual folder/file). It also needs to handle deleting a symlink if the target folder/file no longer exists.
 - Nevermind. I'm going to use `rm -rf` for delete files. It removes a symlink without following. I added unit tests to verify.
- [x] Add unit tests for the minimatch interface change. Madhuri pointed out to me that multiple tasks have a lot of code doing the same thing: applying match patterns. I added an overload so `pattern` can either be a string or an array of strings. I moved the hash table logic (which is used to avoid returning duplicate results) into the lib function.
- [ ] I'm on the fence about this one, but I'm leaning toward switching the find function to the following interface: `find(findPath: string)` and an overload `find(options: string, findPath: string)`. That would make it consistent with the shelljs function interfaces. Then instead of passing a dictionary of options, it would just be a string, e.g. `find('-L', '/some/path')`.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -431,13 +431,13 @@ export function setEnvVar(name: string, val: string): void {
  * @returns   string[]
  */
 export function getDelimitedInput(name: string, delim: string, required?: boolean): string[] {
-    var inval = getInput(name, required);
-    if (!inval) {
+    let inputVal = getInput(name, required);
+    if (!inputVal) {
         return [];
     }
 
-    var result = [];
-    inval.split(delim).forEach(x => {
+    let result: string[] = [];
+    inputVal.split(delim).forEach((x: string) => {
         if (x) {
             result.push(x);
         }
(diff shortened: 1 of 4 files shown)


node/task.ts
```diff
@@ -932,28 +932,88 @@ export function mv(source: string, dest: string, options?: string, continueOnErr
  * @param     findPath     path to find files under
  * @returns   string[]
  */
-export function find(findPath: string): string[] {
+export function find(findPath: string, options?: FindOptions): string[] {
```
ericsciple: Discussed offline with Scott. I need to investigate:
>In find(), can we get into an infinite loop?  Like, what if I have a symlink to my parent folder?

bryanmacfarlane:
Looks good.

As discussed, options is good as a class or interface.  More appropriate for javascript.

Thanks for adding tests

bryanmacfarlane:
This also means the affected task will have to adopt the latest task-lib.  There's a few breaking changes when adopting newest task lib.  Send me review on that one.

Issue #112, opened 2016-08-11
ToolRunner.exec does not work on Windows when both toolPath and arguments contain spaces


adriaanthomas:
Use case and traces: Microsoft/vsts-tasks#2253

It looks like this is caused by nodejs/node#7367.

Manually patching the build agent fixes this (javascript resulting from https://github.com/Microsoft/vsts-task-lib/blob/master/node/toolrunner.ts#L232, slightly rewritten to remove typescript-specific syntax):

``` typescript
var isWindowsShellScript = this.toolPath.endsWith('.cmd') || this.toolPath.endsWith('.bat');
var toolPath = this.toolPath.indexOf(' ') === -1 ? this.toolPath : '"' + this.toolPath + '"';
var args = this.args.map((arg: string) => arg.indexOf(' ') === -1 ? arg : '"' + arg + '"');
var cp = child.spawn(toolPath, args, { cwd: ops.cwd, env: ops.env, shell: isWindowsShellScript });
```

(clearly wrapping args in double quotes is a hack, this is just to prove the point - without that `Gulpfile.js` still is not found).

bryanmacfarlane:
Can you repro an issue with the task-lib with lib >= 0.8 (we're currently on 0.9.x)

bryanmacfarlane:
I confirmed the task-lib >= 0.8 does not have the issue.  The issue is the gulp task is shipping 0.7 (old) version of the task lib.

```javascript
> var tl = require('vsts-task-lib/task');
##vso[task.debug]agent.workFolder=undefined
##vso[task.debug]loading inputs and endpoints
##vso[task.debug]loaded 0
undefined
> var tr = tl.tool('gulp');
undefined
> tr.arg('/path/with spaces/in/it');
> tr.args
[ '/path/with spaces/in/it' ]
> 
```

The key is the path intact with spaces needs to get into process argv with spaces and without double quotes.  We will update the gulp task to use 0.9

adriaanthomas:
You're faster than I am. Does your Gulp path also have spaces? Because that is the only way to fully reproduce this. Your test code only covers the argument, not the path that is passed to `child_process.spawn` as the first argument.

adriaanthomas:
To be specific, I don't see this error addressed yet:

```
2016-08-11T07:06:43.3600568Z 'C:\Program' is not recognized as an internal or external command,
2016-08-11T07:06:43.3600568Z operable program or batch file.
```

[...]

PR #113, opened 2016-08-11
Fix spaces on Windows in cmd and args

adriaanthomas:
Only as a suggestion for #112.

Note: I had to increase the timeouts on Windows, but left them as-is because this probably works better on your end.

Also, it took me a while to realise this, but node.js tasks run in an embedded node, which at this time is 5.10.1. Quoting `toolName` does not work on node 4.4.7.

And apologies for the whitespace changes, default editor settings...


ericsciple:
Fixed as part of PR https://github.com/Microsoft/vsts-task-lib/pull/157. Effectively runs shell=true for .cmd and .bat files and applies different quoting rules.

PR #114, opened 2016-08-13
Test using specific node version.

ericsciple:


node/make.js
```diff
@@ -59,7 +52,152 @@ target.loc = function() {
...
 
-    // Create the en-US resjson file.
+    // create the en-US resjson file.
     var enContents = JSON.stringify(strings, null, 2);
     fs.writeFileSync(path.join(strPath, 'resources.resjson'), enContents)
-}
\ No newline at end of file
+}
+
+var rp = function (relPath) {
```
ericsciple: I moved the `rp` and `run` helper functions down. I had to add a few more helper functions. It seemed like the target.___ assignments were more important to see at the top of the file (rather than leading the file with the helper functions). Also it seemed like all of the helper functions should be grouped together.

node/make.js
```diff
@@ -59,7 +52,152 @@ target.loc = function() {
...
+    else {
+        separator = ':';
+    }
+
+    var existing = process.env['PATH'];
+    if (existing) {
+        process.env['PATH'] = directory + separator + existing;
+    }
+    else {
+        process.env['PATH'] = directory;
```
ericsciple: Trailing or leading ":" on nix means to add the current directory to the PATH (wherever current directory is at the time a cmd is invoked). It is also viewed as a potential security vulnerability - so I took precaution to avoid adding a trailing ":" to the PATH.

node/test/tasklib.ts
```diff
@@ -1268,22 +1278,21 @@ describe('Test vsts-task-lib', function () {
                 errStream: _nullTestStream
             }
 
+            var tool;
```
ericsciple: This test was busted on Ubuntu 14.

node/test/tasklib.ts
```diff
@@ -1857,7 +1866,7 @@ describe('Test vsts-task-lib', function () {
             var jsonPath = path.join(tempFolder, 'task.json');
             fs.writeFileSync(jsonPath, jsonStr);
 
-            var tempLocFolder = path.join(tempFolder, 'strings', 'resources.resjson', 'en-US');
+            var tempLocFolder = path.join(tempFolder, 'Strings', 'resources.resjson', 'en-US');
```
ericsciple: This test was busted on Linux.

node/make.js
```diff
@@ -59,7 +52,150 @@ target.loc = function() {
...
+    switch (platform) {
+        case 'darwin':
+            var nodeArchivePath = downloadArchive(nodeUrl + '/' + nodeVersion + '/node-' + nodeVersion + '-darwin-x64.tar.gz');
+            addPath(path.join(nodeArchivePath, 'node-' + nodeVersion + '-darwin-x64', 'bin'));
+            break;
+        case 'linux':
+            var nodeArchivePath = downloadArchive(nodeUrl + '/' + nodeVersion + '/node-' + nodeVersion + '-linux-x64.tar.gz');
+            addPath(path.join(nodeArchivePath, 'node-' + nodeVersion + '-linux-x64', 'bin'));
+            break;
+        case 'win32':
```
ericsciple: Windows needs an extra step to copy node.exe and node.lib into a directory that can be added to the PATH. On OSX/Linux, the extra step isn't needed because we download and extract a targz. On Windows, the individual files are downloaded like this:
```
C:\vsts-task-lib\node\_download\file\https___nodejs.org_dist_v5.10.1_win-x64_node.exe
C:\vsts-task-lib\node\_download\file\https___nodejs.org_dist_v5.10.1_win-x64_node.lib
```

node/make.js
```diff
@@ -59,7 +52,150 @@ target.loc = function() {
...
+
+var run = function(cl) {
+    console.log('> ' + cl);
+    var rc = exec(cl).code;
+    if (rc !== 0) {
+        echo('Exec failed with rc ' + rc);
+        exit(rc);
+    }
+}
+
```
ericsciple: move to externals.js

PR #116, opened 2016-08-18
Bug in find files. Missing attributes.

ericsciple:


ericsciple:
This is a quick fix. I want to make another fix to add resiliency.

IIRC casting an int to an enum in .Net does not blow up if the value isn't defined on the enum. I did not expect this line to blow up in PowerShell. After this quick fix, I'll continue looking into what can be done for resiliency.

PR #126, opened 2016-08-29
Enable localization for common code modules

jmyersmsft:
Deprecates `setResourcePath` and `loc`, which use a single global string
table, in favor of a new function `loadModuleResources`, which returns a
separate string table for each task.json/module.json passed in.


ericsciple:
Whats wrong with using a global table and prefix to avoid collision?

jmyersmsft:
Global state is kind of icky in general. Yes, there's global state in this PR, but other than the back-compat with setResourcePath it's not really observable unless you're doing unnatural things like editing the resource files on the fly,

Prefixes require some amount of coordination and cooperation, whereas this solution requires none. 

Is there a real use case for consuming a string defined in a separate module? (and by module I mean something of the scope of vsts-task-lib, or our nuget-task-common folder in vsts-tasks)

bryanmacfarlane:
I'll have time soon to look over ...

PR #131, opened 2016-08-31
Fix rmRF for symlink with missing target.

ericsciple:


node/test/dirtests.ts
```diff
@@ -25,6 +25,17 @@ describe('Test Dir Operations', function () {
 
     });
 
+    // this test verifies the expected version of node is being used to run the tests.
```
ericsciple: I just moved this to the top of the file. I don't see a good file for this test so this file suffices i guess.

PR #134, opened 2016-09-02
Expose MatchOptions instead of minimatch.IOptions

ericsciple:


node/task.ts
```diff
@@ -1268,7 +1268,7 @@ export function match(list: string[], pattern: any, options?: MatchOptions): str
     return list.filter((item: string) => map.hasOwnProperty(item));
 }
 
-export function filter(pattern: string, options?: minimatch.IOptions): (element: string, indexed: number, array: string[]) => boolean {
+export function filter(pattern: string, options?: MatchOptions): (element: string, indexed: number, array: string[]) => boolean {
```
scottdallamura: MatchOptions just extends minimatch.IOptions - does this work?

bryanmacfarlane: I just did a checkin where I redefine the options.  The issue is in typings, if someone hasn't done a typings impl (only definatelytyped for example), then consumers of your typings need to install all your dependencies that leak.  They said they are addressing in 2.0.  Obviously I wouldn't redefine a whole module but where some simple options interface leaks, easy enough

PR #137, opened 2016-09-07
Add abstracted Find Files functionality

nigurr:
- Multiple filters can be passed and should be sent as array
- Include/Exclude filter support
- Search files anywhere

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -10,6 +10,8 @@ import tcm = require('./taskcommand');
 import trm = require('./toolrunner');
 import vm = require('./vault');
 import semver = require('semver');
+
+let loadash = require('lodash');
 require('./extensions');
 
 export enum TaskResult {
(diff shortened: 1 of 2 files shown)


node/task.ts
```diff
@@ -1157,6 +1159,27 @@ export function globFirst(pattern: string): string {
     return matches[0];
 }
 
+export function findFiles(includePattern: string[], excludePattern: string[], workingDir?: string): string[] {
```
ericsciple: does this use minimatch pattern matching rules?

ericsciple: If exclude patterns is to be pulled into the lib, then tl.match should have an overload. At that point, this function would become a thin wrapper over tl.find and tl.match.

I like the idea of a wrapper function so caller doesn't need to setup the options (specify matchbase:true and if-windows case-insensitive:true). However, I recently added fixes and unit tests for find and the function you are adding here does not expose the same options (and likely behaves differently too). We should drive consistency, not add different find/globbing functionality.

nigurr: Can you point me to the tests which shows the existing pattern? 
I looked at from Task owners perspective, they expect to simply pass VSTS filters and expects the files to be discovered.

nigurr: @ericsciple ping. We should close on this asap, as we have too much dependency on this functionality across different tasks :)

bryanmacfarlane: We won't be able to this week.  Eric is OOF this week and I'm OOF till Monday.

ericsciple: i still have the same questions and concerns

bryanmacfarlane: Agreed on Eric's points above.

bryanmacfarlane:
In addition to Eric's points, we need to also extend the mock class in the lib with the changes.

ericsciple:
closing since `findMatch` has been added now which addresses other issues and `match` has been revamped

PR #141, opened 2016-09-12
Allow piping output of exec to another tool

madhurig:


node/mock-toolrunner.ts
```diff
@@ -210,7 +216,17 @@ export class ToolRunner extends events.EventEmitter {
...
+                    var pipeToolArgString = this.pipeOutputToTool.args.join(' ') || '';
+                    var pipeToolCmdString = this.ignoreTempPath(this.pipeOutputToTool.toolPath);
+                    if(pipeToolArgString) {
+                        pipeToolCmdString += (' ' + pipeToolArgString);
+                    }
+
+                    cmdString += ' | ' + pipeToolCmdString;
+                }
+
+                ops.outStream.write('[command]' + cmdString + os.EOL);
```
bryanmacfarlane: This should be ops.outStream.write(res.stdout + os.EOL)?
That's the redline to the left.

node/test/toolrunnertests.ts
```diff
@@ -410,6 +410,225 @@ describe('Toolrunner Tests', function () {
...
+            find.arg('"cmd"');
+
+            var tasklist = tl.tool(tl.which('tasklist', true));
+            tasklist.pipeExecOutputToTool(find);
+
+            tasklist.on('stdout', (data) => {
+                output += data.toString();
+            });
+
+            tasklist.exec(_testExecOptions)
```
bryanmacfarlane: can we use await?  this seems so 2015 :))

node/test/toolrunnertests.ts
```diff
@@ -410,6 +410,225 @@ describe('Toolrunner Tests', function () {
...
+                done();
+            })
+
+        }
+        else {
+            var grep = tl.tool(tl.which('grep', true));
+            grep.arg('ssh');
+
+            var ps = tl.tool(tl.which('ps', true));
+            ps.arg('bad');
```
bryanmacfarlane: Can we flatten the if / else code to one shared block of logic?  Seems like we could have variables above on tool names and expected output which are conditionalized on platform and one block of code.

node/mock-toolrunner.ts
madhurig: Good catch, this is in the mock, will fix.

PR #143, opened 2016-09-15
Users/mg/secretdataparams

madhurig:


node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vsts-task-lib",
-  "version": "0.9.19",
+  "version": "0.9.20",
```
bryanmacfarlane: note I just published 20 with another bug fix.  need to bump to 21

madhurig: Will do. Let me know your feedback on the other changes.

bryanmacfarlane:
Look at the ps3 lib.  I seem to remember the ability of the lib consumer (task author) to register what they know to be a secret.  I'll look at it closer tomorrow.  We might want to wait for Eric's feedback

node/taskcommand.ts
```diff
@@ -33,12 +34,16 @@ export class TaskCommand {
...
+                            containsSecret = true;
+                        }
                     }                    
                 }
             }
         }
 
-        cmdStr += ']' + this.message;
+        cmdStr += ']';
+        cmdStr += containsSecret? '********' : this.message;
```
bryanmacfarlane: Not sure I like baking in the concept of a 'secret' property into the command class.  This is a generic parser.

Why is this change needed?  When is the toString of the cmd leaking to a log?  And why wouldn't registering it as a secret suffice via setVariable?

bryanmacfarlane: OK - I see why because you call cmd and it debug logs cmd tostring.  So once again you're trying to register the secret but the act of registering the secret is logging the secret.

madhurig: SetVariable() calls 
command('task.setvariable', { 'variable': name || '', 'secret': (secret || false).toString() }, varValue) which write out the value to the console/log even if it is a secret. I fixed it in the toString() but we could do it in the command() function as well: https://github.com/Microsoft/vsts-task-lib/blob/master/node/task.ts#L618

madhurig: Actually setVariable() is broken if secret is true. If we don't mask in the command, the value is written without being masked. If I mask it, the command actually does not get the value i.e. the variable is not set for future use during the build. In my case the fix works since I don't use the variable again but the fix is not correct.

node/task.ts
```diff
@@ -516,15 +516,22 @@ export function getEndpointUrl(id: string, optional: boolean): string {
...
 
     if(!optional && !dataParamVal) {
         throw new Error(loc('LIB_EndpointDataNotExist', id, key));
     }
 
+    if(secret) {
+        setVariable(dataVariable, dataParamVal, true);
+        dataParamVal = vault.retrieveSecret('SECRET_' + dataVariable);
+    }
+
```
bryanmacfarlane: // for RC2.  In RTM after 2.x has use command setSecret
if (secret) {
   debug(id + ' data ' + key + ' hasValue ' + dataParamVal ? true : false);
}
else {
    debug(id + ' data ' + key + ' = ' + dataParamVal);
}

madhurig:
Abandoning this change for RC2.

PR #144, opened 2016-09-16
avoid writing value of secret endpoint data to debug output

madhurig:


madhurig:
doing a more localized fix in the task

PR #145, opened 2016-09-23
Upgrade TS to 2.x

nigurr:


bryanmacfarlane:
We will do this but can't evaluate till sometime next week.  Churning the task repo this week ...

node/package.json
```diff
@@ -31,6 +31,7 @@
     "q": "^1.1.2",
     "semver": "^5.1.0",
     "shelljs": "^0.3.0",
+    "typescript": "^2.0.3",
```
bryanmacfarlane: This is a dev dependency

nigurr: Awww. Fixed it.

bryanmacfarlane:
Let's hold for 108.  I want to make sure 107 stabilizes and we consciously move to 2.0.  I will send communications to teams inline with Trevor's mail

bryanmacfarlane:
since we started a 2.x preview task lib, we can consider taking this.  we have to carefully look at all the 2.0 implications but it's a good time.  The doc gen might also be impacted since we load the ts compiler for that.

stephenmichaelf:
This has been done in a separate PR. Closing.

PR #146, opened 2016-09-25
Add task.json schema file

AArnott:
Fixes Microsoft/vsts-tasks#1854

### DIFF
--- a/tasks.schema.json
+++ b/tasks.schema.json
@@ -1,0 +1,335 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "title": "VSTS Tasks schema",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "A unique guid for this task",
+      "pattern": "^[A-Fa-f0-9]{8}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{12}$"
+    },
+    "name": {
+      "type": "string",
+      "description": "Name with no spaces",
+      "pattern": "^[A-Za-z0-9\\-]+$"
+    },
+    "friendlyName": {
+      "type": "string",
+      "description": "Descriptive name (spaces allowed)"
+    },
+    "description": {
+      "type": "string",
+      "description": "Detailed description of what your task does"
+    },
+    "helpMarkDown": {
+      "type": "string"
+    },
+    "author": {
+      "type": "string"
+    },
+    "visibility": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "enum": [
+          "Build",
+          "Release",
+          "Preview"
+        ]
+      }
+    },
+    "runsOn": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "enum": [
+          "Agent",
+          "MachineGroup",
+          "Server"
+        ]
+      }
+    },
+    "category": {
+      "type": "string",
+      "description": "Where the task appears in VSTS",
+      "enum": [
+        "Build",
+        "Utility",
+        "Test",
+        "Package",
+        "Deploy"
+      ]
+    },
+    "groups": {
+      "type": "array",
+      "description": "Describes groups that task properties may be logically grouped by in the UI.",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": [
+          "name",
+          "displayName"
+        ],
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "displayName": {
+            "type": "string"
+          },
+          "isExpanded": {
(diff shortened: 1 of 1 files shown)


bryanmacfarlane:
Eric and Yash, can you review the schema file?

ericsciple:
@bryanmacfarlane do you have any thoughts where the file should go? I'm guessing once we choose a location, we need to leave it there.

This is nice.

ericsciple:
@AArnott fyi - i haven't forgotten about this one... lot going on right now, shutting down for RTM and normal 3wk sprints going on in parallel.

ericsciple:
@bryanmacfarlane are you fine with the location of this file (root of the repo)? i'm guessing once we choose a location it needs to stay there (since folks will be hardcoding a dependency to that specific url)

tasks.schema.json
```diff
@@ -0,0 +1,335 @@
...
+              "connectedService:ssh",
+              "connectedService:Generic",
+              "connectedService:Jenkins",
+              "connectedService:servicefabric",
+              "filePath",
+              "multiLine",
+              "pickList",
+              "radio",
+              "string"
+            ]
```
ILMTitan: Because you can add new "connectedService:" within an extension, I thing this enum needs to be:

            "anyOf": [
              {
                "enum": [
                  "boolean",
                  "filePath",
                  "multiLine",
                  "pickList",
                  "radio",
                  "string"
                ]
              },
              {
                "pattern": "^connectedService\\:.*$"
              }
            ]

This still gives the intellisense for the standard inputs, but also avoids validation errors for new connectedServices. Adding known connectedServices to the enum would also be reasonable.

AArnott: Good idea. I'd fix this, but would like to know if this PR isn't just dead first. So far, it seems to be that no one with commit privileges is acting on it.

ILMTitan: Which is silly, considering the [official Microsoft documentation](https://www.visualstudio.com/en-us/docs/integrate/extensions/develop/build-task-schema) still references [your schema](https://raw.githubusercontent.com/AArnott/vso-agent-tasks/9b41d803a/tasks.schema.json).

ericsciple:
@AArnott The main concern I have is, once it is merged, then is the location sealed? Or would an fwlink work in the `$schema` property? Would that prevent folks from breaking if the location is changed?

AArnott:
I suspect json language services can handle HTTP redirects, but I haven't tried it. FWIW it looks like folks are *already* burning in a specific commit's version of this schema, so by completing the PR and updating the docs we at least get them to 'tip' of whatever you have. But assuming fwlink's work (or perhaps since this is customer visible an aka.ms link would be better) are you ok with merging this then?

AArnott:
OK, I just tested it. VS Code (for one) supports redirects in their JSON schema URLs. I set up https://aka.ms/vsts-tasks.schema.json and made you, @ericsciple, co-owner of it so you can update it once this PR completes.

ericsciple:
OK my last concern is how do folks know they should use the aka link, and not whatever URL they happen to be browsing when they come across the schema document. It looks like json schema supports an "id" property that can be used to identify the URL of the schema document itself, e.g.
```json
{
    "id": "https://aka.ms/vsts-tasks.schema.json"
```

So that addresses my last concern.

ericsciple:
@bryanmacfarlane any outstanding concerns? I can add the ID at the top after merged, or whatever.

AArnott:
Ping. I'm seeing random folks on github referencing this .json file from my fork. The sooner this PR gets completed (or guidance for placing it elsewhere is given), the sooner folks can reference the official location and it can be properly maintained.

AArnott:
Is this the PR [you were referring to](https://github.com/Microsoft/vsts-tasks/pull/1066#issuecomment-342560110), @bryanmacfarlane?

tasks.schema.json
```diff
@@ -0,0 +1,334 @@
...
+  "definitions": {
+    "executionObject": {
+      "type": "object",
+      "additionalProperties": true,
+      "properties": {
+        "target": {
+          "type": "string",
+          "description": "The target file to be executed. You can use variables here in brackets e.g. $(currentDirectory)\filename.ps1"
+        },
+        "argumentFormat": {
```
masters3d: @bryanmacfarlane Is there a description for argumentFormat? I've searched but have not found what argument format is suppose to do. For most examples is just an empty string.  Fox example: Can I pass arguments here?

ericsciple: It is from legacy handlers and it should not be used.

ILMTitan:
The id <https://aka.ms/vsts-tasks.schema.json> currently redirects to the wrong place. It points at <https://github.com/Microsoft/vsts-task-lib/blob/master/tasks.schema.json> but needs to point at <https://raw.githubusercontent.com/Microsoft/vsts-task-lib/master/tasks.schema.json>

ericsciple:
thanks, will fix

PR #151, opened 2016-10-03
Handling for legacy agent

Ajay-MS:


node/task.ts
```diff
@@ -457,7 +457,7 @@ export function getDelimitedInput(name: string, delim: string, required?: boolea
 export function filePathSupplied(name: string): boolean {
     // normalize paths
     var pathValue = this.resolve(this.getPathInput(name) || '');
-    var repoRoot = this.resolve(this.getVariable('build.sourcesDirectory') || '');
+    var repoRoot = this.resolve(this.getVariable('build.sourcesDirectory') || this.getVariable('System.DefaultWorkingDirectory') || '');
```
bryanmacfarlane: Just for code consistency, make this system.defaultWorkingDirectory (casing).  It ultimately doesn't matter since it's all upper cased but looks more consistent.  Also break it on another line after the || and align it.

bryanmacfarlane:
Why do you say just legacy agent?  I would expect it not to work in RM even with the new agent since repo root is not set in RM???

bryanmacfarlane:
BTW, thanks for fixing this.

node/task.ts
TingluoHuang: I thought there is a variable that means differently between Build and RM, should we just use that one?

Ajay-MS:
I mentioned legacy only here because since in case of release it tries to resolve empty string 
`this.resolve('') `
In case of coreclr (2.x)  it is resolving to current working directory so there is no issue. But in case of legacy( 1.x ) agent it empty string is resolving to task directory which is causing error.

PR #157, opened 2016-10-14
Windows verbatim arguments and .cmd quoting rules

ericsciple:


ericsciple:
Add test for tool is .cmd with space in the file path, and tool is .exe with space in path. Based on skimming forums, the file path may need to be quoted manually when using the verbatim option. Although based on >=4.x source code it doesn't look like it.

node/toolrunner.ts
```diff
@@ -263,8 +266,16 @@ export class ToolRunner extends events.EventEmitter {
             // https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options
 
             //start the child process for both tools
-            var cpFirst = child.spawn(toolPathFirst, this.args, {cwd: ops.cwd, env: ops.env});
-            cp = child.spawn(toolPath, this.pipeOutputToTool.args, {cwd: ops.cwd, env: ops.env});
+            var spawnOptionsFirst = { } as child.SpawnOptions;
+            spawnOptionsFirst.cwd = ops.cwd;
+            spawnOptionsFirst.env = ops.env;
+            spawnOptionsFirst['windowsVerbatimArguments'] = ops.windowsVerbatimArguments;
```
bryanmacfarlane: should we only set this if it's true?  might not matter.  since undocumented - if node ever removed it and validated unknown options, then only tasks that used this minor case option would be affected instead of all.  The flip side of the argument is we would find out sooner after adopting a new node.

bryanmacfarlane:
have we reproduced the gulp error and confirmed this does in fact fix it?

ericsciple:
this doesn't fix the gulp issue.

i think this fix will be part of the fix for the gulp issue. i've spent a ton of time playing with the win32 api and reading node source code. i can repro the issue when calling the win32 api directly. i think i know of a fix, but need to spend some time fitting it into a clean consumer-friendly interface.

now that i also have a good grasp on the gulp issue. i think we need to hold this PR for that fix as well.

node/test/toolrunnertests.ts
```diff
@@ -34,15 +36,15 @@ describe('Toolrunner Tests', function () {
 
         tl.pushd(__dirname);
 
-        var _testExecOptions: trm.IExecOptions = {
+        var _testExecOptions = <trm.IExecOptions>{
```
ericsciple: I added another field to the interface, which causes compile errors. Rather than explicitly specify the field everywhere this object is constructed, I opted to just chang how the object is instantiated so we don't need to worry about this next time a field is added.

bryanmacfarlane:
Looks good.  thanks for adding all the tests

node/toolrunner.ts
```diff
@@ -127,6 +121,338 @@ export class ToolRunner extends events.EventEmitter {
...
+        let result: IExecOptions = <IExecOptions>{
+            cwd: options.cwd || process.cwd(),
+            env: options.env || process.env,
+            silent: options.silent || false,
+            failOnStdErr: options.failOnStdErr || false,
+            ignoreReturnCode: options.ignoreReturnCode || false
+        };
+        result.outStream = options.outStream || <stream.Writable>process.stdout;
+        result.errStream = options.errStream || <stream.Writable>process.stderr;
+        result['windowsVerbatimArguments'] = options.windowsVerbatimArguments || false;
```
TingluoHuang: result.windows... = ...

node/toolrunner.ts
```diff
@@ -127,6 +121,338 @@ export class ToolRunner extends events.EventEmitter {
...
+
+        // append second tool
+        if (this.pipeOutputToTool) {
+            cmd += ' | ' + this.pipeOutputToTool._getCommandString(options, /*noPrefix:*/true);
+        }
+
+        return cmd;
+    }
+
+    private _getToolPath(options: IExecOptions): string {
```
TingluoHuang: _getspwnToolPath

node/toolrunner.ts
```diff
@@ -127,6 +121,338 @@ export class ToolRunner extends events.EventEmitter {
...
+        if (process.platform == 'win32') {
+             if (this._isCmdFile()) {
+                 return process.env['COMSPEC'] || 'cmd.exe';
+             }
+        }
+
+        return this.toolPath;
+    }
+
+    private _getArgs(options: IExecOptions): string[] {
```
TingluoHuang: _getspwnArgs

node/test/toolrunnertests.ts
```diff
@@ -720,4 +723,802 @@ describe('Toolrunner Tests', function () {
...
+            assert.equal(tr._windowsQuoteCmdArg('hello`world'), 'hello`world');
+            assert.equal(tr._windowsQuoteCmdArg('hello~world'), 'hello~world');
+            assert.equal(tr._windowsQuoteCmdArg('hello|world'), 'hello|world');
+            assert.equal(tr._windowsQuoteCmdArg('hello<world'), 'hello<world');
+            assert.equal(tr._windowsQuoteCmdArg('hello>world'), 'hello>world');
+
+            done();
+        });
+
+        it('_windowsQuoteCmdArg quotes .cmd args (Windows)', function (done) {
```
TingluoHuang: .bat

PR #158, opened 2016-10-15
1) Tests pass for the wrong reason. 2) Fix intermittent failure in lo…

ericsciple:
…c tests.


ericsciple:
This is going to leave a couple tests busted (pipe between tool runner). I'm not sure whether it's a real issue or a test issue.

Tests were passing before because `done()` was always getting called and `done(err)` should be called for failure cases.

PR #160, opened 2016-10-24
Fix Ubuntu tests

ericsciple:


ericsciple:
This only fixes two of the failed tests.

node/test/toolrunnertests.ts
```diff
@@ -294,16 +294,18 @@ describe('Toolrunner Tests', function () {
                 });
         }
         else {
-            var ls = tl.tool(tl.which('ls', true));
-            ls.arg('-j');
-
+            var bash = tl.tool(tl.which('bash', true));
```
ericsciple: `ls -j` returns 1 on OSX, but 2 on Ubuntu 14.

I just switched to `bash -c` where we have control over stdout, stderr, and exit code.

node/test/toolrunnertests.ts
```diff
@@ -351,7 +353,7 @@ describe('Toolrunner Tests', function () {
     it('Fails on stderr if specified', function (done) {
         this.timeout(1000);
 
-        var scriptPath = path.join(__dirname, 'scripts', 'stderrOutput.js');
+        var scriptPath = path.join(__dirname, 'scripts', 'stderroutput.js');
```
ericsciple: Case sensitivity caused the test to fail Linux.

ericsciple:
reviewed w/ @madhurig offline

PR #161, opened 2016-10-25
Set mock tl function

ericsciple:


node/mock-run.ts
```diff
@@ -18,17 +18,28 @@ export class TaskMockRunner {
...
         mockery.registerMock(modName, mod);
     }
-    
-    public run() {
-        mockery.enable({warnOnUnregistered: false});
-        var tlm = require('vsts-task-lib/mock-task');
-        if (this._answers) {
-            tlm.setAnswers(this._answers);
+
+    public run(noMockTask: boolean) {
```
ericsciple: Also I added a bool here whether to register mock-task. What I'm thinking is, the runner is still useful for L1 scenarios where the tl would not be mocked.

bryanmacfarlane:
This makes sense

node/mock-run.ts
```diff
@@ -18,17 +18,28 @@ export class TaskMockRunner {
         this._answers = answers;
     }
 
+    public setFunction(name: string, val: any) {
```
ericsciple: `registerMockExport`

ericsciple:
this is ready for review now

PR #163, opened 2016-10-27
Allowing arguments to be passed to scripts executed by MockTestRunner

phil-hodgson:


ericsciple:
It looks like this is because `setInput` lives on mock-run, and not mock-test (which is called from the outer process).

Do you think it would make more sense if mock-test exposed an overload that accepted an env dictionary to be used when spawning the child process? That would achieve two things:

1. Avoids arg parsing code in target script. I'm thinking it may be easier (less code) and better for readability (avoids requiring an arg parsing module and/or interpreting positional parameters passed to the process).

2. Avoids quoting issues involved in passing command line args to child processes. When spawning a child process on Windows, Node applies quoting logic that (hopefully) matches up with how the target process parses it's own command line. Since this is for testing, I imagine it might be useful to test corner cases with special characters. If so, then avoiding the quoting/parsing altogether seems good - less likely to run into issues from that extra layer of translation.

By exposing an env dictionary overload, technically you could even just set INPUT_FOO directly from the outer script (without dirty-ing the outer process env). Another option would be to add a `setInput` function to mock-test, but I have concerns that having two `setInput` functions would be confusing for consumers.

phil-hodgson:
I'd be ok with passing a dictionary to the `run` method. Avoiding parameter parsing on the target script's end is appealing, but I'd prefer to have a method to retrieve the "testScriptValues" rather than directly accessing the environment variables. The env vars is an un-scoped bag of "stuff" with no guarantee of avoiding name collisions, so wrapping this access would be cleaner.

damccorm:
Do we still need this? It seems like it could be useful, I think I agree with Eric that passing in a dictionary feels cleaner though.

damccorm:
I'm going to close. We'll look into changing how we do this with the new major version

PR #165, opened 2016-11-01
Allowing unmocked filters to use the base behavior

phil-hodgson:
If you don't mock a filter, then calls to task.filter returns a function that doesn't match anything. This change ensures that an un-mocked filter returns the base behavior.

### DIFF
--- a/node/mock-task.ts
+++ b/node/mock-task.ts
@@ -407,11 +407,13 @@ export function matchFile(list, pattern, options): string[] {
     return mock.getResponse('match', pattern) || [];
 }
 
-export function filter(pattern, options): any {
-    var filterList = mock.getResponse('filter', pattern) || [];
-	return function(pattern, i, arr) {
-		return filterList.indexOf(pattern) >= 0;
-	}
+export function filter(pattern, options): (element: string, index: number, array: string[]) => boolean {
+    var filterList: string[] = mock.getResponse('filter', pattern);
+    if (filterList) {
+        return (element, index, array) => (filterList.indexOf(element) !== -1);
+    }
+    // default back to the built-in behavior
+    return task.filter(pattern, options);
 }
 
 //-----------------------------------------------------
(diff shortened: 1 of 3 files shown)


node/mock-task.ts
```diff
@@ -407,11 +407,13 @@ export function matchFile(list, pattern, options): string[] {
     return mock.getResponse('match', pattern) || [];
 }
 
-export function filter(pattern, options): any {
-    var filterList = mock.getResponse('filter', pattern) || [];
-	return function(pattern, i, arr) {
-		return filterList.indexOf(pattern) >= 0;
-	}
+export function filter(pattern, options): (element: string, index: number, array: string[]) => boolean {
```
ericsciple: On the one hand, I really like the idea of functions like `filter` (and `match`) being able to use the real behavior. However, what feels strange to me about this implementation is, *some* `filter` calls may be mocked, and others use the real functionality.

I think an all-or-nothing behavior would be more inline with common mocking approaches. That is, only "fallback" if 'filter' isn't registered at all in the answer object. That would mean changing MockAnswer to expose a function like `isRegistered(cmd: string): boolean`.

Thgouhts?

phil-hodgson: With Moq mocking (C#) it's possible to mock some calls to a method (based on param values), and have others fall through to the base implementation (if `CallBase=true`), so that part doesn't feel odd to me.

From my test author's point of view the base implementation of 'filter' is perfectly good - generate a function that matches values on a pattern - there's no external dependencies on the environment.

My motivation was that I had a portion of build task code `tl.find(path).filter(tl.filter('*.exe'))`, and I wanted to mock `find` (since I didn't want to rely on the real file system). The current mocking framework forces me to _also_ get involved in mocking `filter` (whose default behavior would have been perfectly good), which then makes my test code unnecessarily messy.

I'd be happy not mocking `filter` at all, but if I do want to, then I don't really want to be forced to mock _every_ occurrence that appears _somewhere_ in the task code. I'd prefer to be targetted with the mocking.

ericsciple:
removed answers support for `match` and `filter` in PR https://github.com/Microsoft/vsts-task-lib/pull/179

those functions dont touch disk so task author already has control over their inputs through other mock functions

PR #166, opened 2016-11-01
Emitting a debug message when an exec call is not mocked

phil-hodgson:
To help with test development, this change writes out details of an exec call that is made without being mocked.

### DIFF
--- a/node/mock-toolrunner.ts
+++ b/node/mock-toolrunner.ts
@@ -222,6 +222,10 @@ export class ToolRunner extends events.EventEmitter {
 
         // TODO: filter process.env
         var res = mock.getResponse('exec', cmdString);
+        if (!res) {
+            // exec call not mocked correctly?
+            this._debug(`'exec' call not mocked: ${cmdString}`);
+        }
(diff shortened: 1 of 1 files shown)


node/mock-toolrunner.ts
```diff
@@ -222,6 +222,10 @@ export class ToolRunner extends events.EventEmitter {
 
         // TODO: filter process.env
         var res = mock.getResponse('exec', cmdString);
+        if (!res) {
+            // exec call not mocked correctly?
+            this._debug(`'exec' call not mocked: ${cmdString}`);
+        }
```
ericsciple: This makes me wonder whether this should be baked into `mock.getResponse()` instead. That seems more appropriate to me and I wouldn't think it would introduce any problem.

ericsciple: @bryanmacfarlane any thought about that?

ericsciple:
fixed in following PR in a generic way https://github.com/Microsoft/vsts-task-lib/pull/179

PR #170, opened 2016-11-04
Added findFiles method for searching file based on regex, it supports…

Ajay-MS:
Since tl.glob method doesn't supports shared path added method for searching files based on pattern that works for shared path as well.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -1170,6 +1170,50 @@ export function globFirst(pattern: string): string {
     return matches[0];
 }
 
+export function findFiles(pattern: string){
+    var filesList : string [];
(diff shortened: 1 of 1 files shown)


ericsciple:
thanks. i dont want to merge this PR, but this is useful to me. i'm working on a find files function right now and solving this problem is one of the goals. there is another current limitation that i want to work into one robust function: support interleaved exclude patterns. also the function needs to accept FindOptions and MatchOptions. and also should support rooting unrooted paths.

ericsciple:
closing since `findMatch` has been added now which addresses other issues

PR #175, opened 2016-11-15
Finding files markdown

ericsciple:


node/README.md
```diff
@@ -15,6 +15,8 @@ Cross platform tasks are written in Typescript.  It is the preferred way to writ
 
 Step by Step: [Create Task](docs/stepbystep.md)  
 
+Scenario Guidance: [Finding Files](docs/findingfiles.md)
```
bryanmacfarlane: I would just call "Guidance".  Put it after the doc link below.

node/docs/findingfiles.md
```diff
@@ -0,0 +1,98 @@
...
+    nocase = process.platform == 'win32' // case insensitive on Windows, otherwise case sensitive
+    nonull = false,
+    matchBase = false,
+    nocomment = false,  // support comments
+    nonegate = false,   // support negate pattern
+    flipNegate = false
+}
+```
+
+## Pattern Syntax
```
ericsciple: I think a separate doc makes sense for user docs. For example, in the user docs we won't mention brace expansion at all because it will be off for in the box tasks.

PR #178, opened 2016-11-25
Internal cleanup

ericsciple:


node/docs/releases.md
```diff
@@ -1,5 +1,8 @@
 # VSTS-TASK-LIB RELEASES
 
+## 2.0.1-preview
```
ericsciple: i have another branch in parallel for other 2.0.1 changes, but wanted to discuss this particular set of changes separately

node/make.js
```diff
@@ -38,6 +38,7 @@ target.test = function() {
     buildutils.getExternals();
     run('tsc -p ./test');
     cp('-Rf', rp('test/scripts'), testPath);
+    process.env['TASKLIB_INPROC_UNITS'] = '1'; // export task-lib internals for internal unit testing
```
ericsciple: i changed it up so `_internal` is only exported if this env var is set, which means it needs to be set before the first `require('task')`. before launching the mocha process is the easiest way to do that.

node/mock-task.ts
```diff
@@ -25,25 +25,25 @@ export enum TaskResult {
...
+let _outStream = process.stdout;
+let _errStream = process.stderr;
 
-export function _writeError(str: string): void {
+function _writeError(str: string): void {
     _errStream.write(str + os.EOL);
 }
 
-export function _writeLine(str: string): void {
+function _writeLine(str: string): void {
```
ericsciple: i changed these to not be exported. i couldn't find any reason why they need to be exported. am i missing something?

node/task.ts
```diff
@@ -17,7 +17,10 @@ export enum TaskResult {
     Failed = 1
 }
 
-export var _internal = {};
+let _internal = { } as any;
+if (process.env.TASKLIB_INPROC_UNITS) {
+    module.exports._internal = _internal;
```
ericsciple: as long as TASKLIB_INPROC_UNITS is not defined during build, then _internal will not get exported and will not end up in the .d.ts

note, this piggybacks on the env var TASKLIB_INPROC_UNITS which is already used for other internal unit test behavior

node/task.ts
```diff
@@ -28,33 +31,33 @@ export var _internal = {};
  *  2) to distinguish between secret vars and public
  *  3) to know the real variable name and not just the formatted env var name.
  */
-let knownVariableMap: { [key: string]: KnownVariableInfo; } = { };
+let _knownVariableMap: { [key: string]: _KnownVariableInfo; } = { };
```
ericsciple: i prefixed all private fields too. seemed like the right thing to do

node/task.ts
```diff
@@ -356,7 +357,7 @@ function getVariableKey(name: string): string {
  *  1) the real variable name (not the formatted environment variable name)
  *  2) whether the variable is a secret variable
  */
-interface KnownVariableInfo {
+interface _KnownVariableInfo {
```
ericsciple: and prefixed private interfaces too

node/task.ts
```diff
@@ -1197,9 +1191,9 @@ function legacyFindFiles_convertPatternToRegExp(pattern: string): RegExp {
     let flags = process.platform == 'win32' ? 'i' : '';
     return new RegExp(pattern, flags);
 }
-legacyFindFiles['legacyFindFiles_convertPatternToRegExp'] = legacyFindFiles_convertPatternToRegExp; // for unit testing
+_internal._legacyFindFiles_convertPatternToRegExp = _legacyFindFiles_convertPatternToRegExp; // for unit testing
```
ericsciple: every internal member that needs to be exposed for the internal unit tests, now hangs off of `_internal`

node/task.ts
```diff
@@ -2030,15 +2024,15 @@ if (semver.lt(process.versions.node, '4.2.0')) {
 // Populate the vault with sensitive data.  Inputs and Endpoints
 //-------------------------------------------------------------------
 
-// only exposed as a function so unit tests can reload vault for each test
+// only exposed as a function so unit tests can reload vault for each test.
 // in prod, it's called globally below so user does not need to call
 
-export function _loadData(): void {
+function _loadData(): void {
```
ericsciple: `_loadData` is for internal unit testing only, correct? otherwise it needs to continue to be exported

node/task.ts
```diff
@@ -402,13 +403,6 @@ export function getBoolInput(name: string, required?: boolean): boolean {
     return (getInput(name, required) || '').toUpperCase() == "TRUE";
 }
 
-// deprecated - use  setVariable
-export function setEnvVar(name: string, val: string): void {
```
ericsciple: seemed like a good time to remove this

TingluoHuang:
looks good.

node/task.ts
```diff
@@ -28,33 +31,33 @@ export var _internal = {};
...
+let _outStream = process.stdout;
+let _errStream = process.stderr;
 
-export function _writeError(str: string): void {
+function _writeError(str: string): void {
     _errStream.write(str + os.EOL);
 }
 
-export function _writeLine(str: string): void {
+function _writeLine(str: string): void {
```
ericsciple: doc in readme, use console.log

PR #181, opened 2016-11-29
Replace space in variable key with underscore

ericsciple:


ericsciple:
matches agent: https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Worker/Handlers/Handler.cs#L138

TingluoHuang:
Looks good.

PR #188, opened 2016-12-17
Add support for SucceededWithIssues to node sdk

jessehouwing:
Adds SucceededWithIssues to the list of possible task outcomes to be in line with the PowerShell SDK.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/vsts-task-lib/188)

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -14,6 +14,7 @@ require('./extensions');
 
 export enum TaskResult {
     Succeeded = 0,
+    SucceededWithIssues = 2,
     Failed = 1
 }
 
@@ -79,7 +80,7 @@ export function setErrStream(errStream): void {
  * If not set, task will be Succeeded.
  * If multiple calls are made to setResult the most pessimistic call wins (Failed) regardless of the order of calls.
  * 
- * @param result    TaskResult enum of Success or Failed.  
+ * @param result    TaskResult enum of Succeeded, SucceededWithIssues or Failed.  
  * @param message   A message which will be logged as an error issue if the result is Failed.
  * @returns         void
  */
@@ -89,6 +90,9 @@ export function setResult(result: TaskResult, message: string): void {
     // add an error issue
     if (result == TaskResult.Failed && message) {
         error(message);
+    }
+    else if (result == TaskResult.SucceededWithIssues && message) {
+        warning(message);
     }
 
     // set the task result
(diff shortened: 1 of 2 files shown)


node/task.ts
```diff
@@ -79,7 +80,7 @@ export function setErrStream(errStream): void {
  * If not set, task will be Succeeded.
  * If multiple calls are made to setResult the most pessimistic call wins (Failed) regardless of the order of calls.
  * 
- * @param result    TaskResult enum of Success or Failed.  
+ * @param result    TaskResult enum of Succeeded, SucceededWithIssues or Failed.
```
ericsciple: the verbiage on the message param is out of sync now

node/task.ts
```diff
@@ -14,6 +14,7 @@ require('./extensions');
 
 export enum TaskResult {
     Succeeded = 0,
+    SucceededWithIssues = 2,
```
ericsciple: `node/mock-task.ts` also needs the same change.

i dont see anywhere that parses a number back to the enum. the value is serialized using the enum name in `##` command. i'd rather change the order and keep this clean (succeeded=0, succeededwithissues=1, failed=2)

node/task.ts
```diff
@@ -90,6 +91,9 @@ export function setResult(result: TaskResult, message: string): void {
     if (result == TaskResult.Failed && message) {
         error(message);
     }
+    else if (result == TaskResult.SucceededWithIssues && message) {
+        warning(message);
+    }
```
ericsciple: i think this is appropriate. the user chooses whether to pass a message param.

node/task.ts
jessehouwing: Agreed, but I'm wondering what it will do to tasks that use structures like this:

https://github.com/Microsoft/vsts-tasks/blob/f3bcf5ff1382a435505df1f5b76851ecc850f671/Tasks/Npm/npmtask.ts#L37

It's why I'm not sure what value to give here to be honest...

node/task.ts
jessehouwing: Will update

node/task.ts
ericsciple: that line will always pass zero (otherwise will transfer control to the catch block, which uses the enum)

node/task.ts
jessehouwing: Seems it was out of sync before. Now it matches the enum.

jessehouwing:
Merge?

ericsciple:
fyi - i've got a couple small fixes i'm working on now and a new version of the lib should be published shortly after that.

jessehouwing:
Nice!

PR #200, opened 2017-01-27
Fix String.prototype.EndsWith to be ES6 compliance

ronwang01:
Fix the String.prototype.endsWith polyfill to unblock the issue with Node unable to complete TLS handshake.


ericsciple:
Thanks for the heads up on this issue. I removed the prototype functions (PR https://github.com/Microsoft/vsts-task-lib/pull/207) so closing this PR

PR #213, opened 2017-03-08
Add ability to mock external tools with a function.

ILMTitan:
I had also considered not adding execTool and adding the function type a type union on exec.

### DIFF
--- a/node/test/mocktests.ts
+++ b/node/test/mocktests.ts
@@ -183,5 +183,56 @@ describe('Mock Tests', function () {
         
         assert(tool, "tool should not be null");
         assert(rc == 0, "rc is 0");
-    })                
+    })
+
+    it('Mock toolRunner returns success code for execTool', async () => {
+        var a: ma.TaskLibAnswers = <ma.TaskLibAnswers>{
+            "execTool": {
+                "/usr/local/bin/atool": (args: string[]) => {
+		    assert.equal(args.indexOf("--arg"), 0, "--arg should be the first element");
+		    assert.equal(args.indexOf("foo"), 1, "foo should be the second element");
+		    return {
+			"code": 0,
+			"stdout": "atool output here",
+			"stderr": "atool with this stderr output"
+                    }
+		}
+            }
+        };
+
+        mt.setAnswers(a);
+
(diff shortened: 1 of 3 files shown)


node/test/mocktests.ts
```diff
@@ -183,5 +183,56 @@ describe('Mock Tests', function () {
...
         assert(rc == 0, "rc is 0");
-    })                
+    })
+
+    it('Mock toolRunner returns success code for execTool', async () => {
+        var a: ma.TaskLibAnswers = <ma.TaskLibAnswers>{
+            "execTool": {
+                "/usr/local/bin/atool": (args: string[]) => {
+		    assert.equal(args.indexOf("--arg"), 0, "--arg should be the first element");
+		    assert.equal(args.indexOf("foo"), 1, "foo should be the second element");
```
ericsciple: I see. If you want to plug in your own function, another way is with registerMockExport. Originally the answers came from a file, so the answers interfaces is currently simple types.

ericsciple: [example here](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/CopyFiles/Tests/L0cleansIfSpecified.ts#L28)

ILMTitan: Your example registers a mock for task.stats, does it not? I want to register a mock for the tool being run by toolrunner. Currently you have to define the whole exec string


```
const execString = '/mocked/tools/tool' + ' tool command' +
    ` "${positionArg}"` + ` --arg1="${arg1Val}"` +
    ' --project=projectId' + ' --quiet' + ' --verbosity=info' +
    ` --version="${versionVal}"` + ' --promote' + ' --stop-previous-version';

answers.exec[execString] = {
    "code": 1,
    "stdout": '[tool output]',
    "stderr": '[tool error]'
};
```

I would like to make that easier by having an answer that only needs to know the tool path and can check that the args are correct.

```
answers.execTool['/mocked/tools/tool'] = (args: string[]) => {
    return { code: 1, stderr: '[tool error]'}
}
```

damccorm:
@ILMTitan are you still interested in reviving this? Not 100% sure we need this, but I'd be open to taking a look if you're still interested.

ILMTitan:
I am no longer working on the project that prompted this.

PR #217, opened 2017-03-13
add setState() and getState() to support wrapper task.

ericsciple:


node/task.ts
```diff
@@ -656,13 +656,52 @@ export function getSecureFileName(id: string): string {
...
+//-----------------------------------------------------
+// Task State Helpers
+//-----------------------------------------------------
+/**
+ * Gets a state value that is set by previous step of the same wrapper task.
+ * 
+ * @param     name     name of the state to get
+ * @returns   string
+ */
+export function getState(name: string): string {
```
ericsciple: "getTaskState"? or "getTaskVariable"?

ericsciple: getTaskVariable

PR #232, opened 2017-03-31
Prevent _loadData() from being called twice

ericsciple:


node/task.ts
```diff
@@ -2221,9 +2221,16 @@ function _loadData(): void {
...
+    // avoid loading twice (overwrites .taskkey)
+    global['_vsts_task_lib_loaded'] = true;
 }
 _internal._loadData = _loadData;
-_loadData();
+
+// avoid loading twice (overwrites .taskkey)
+if (!global['_vsts_task_lib_loaded']) {
+    _loadData();
+}
```
ericsciple: When the task lib is loaded, we yank sensitive info out of the environment block and store it encrypted in memory. Normally require caches the module, so even if another file calls require on the task lib, _loadData() won't get called a second time. If _loadData is called a second time, then a a new symmetric key is generated and previously loaded inputs become garbled.

The problem is, the caching provided by `require` doesn't solve every scenario. For instance, the vsts-task-tool-lib carries a specific version of the vsts-task-lib. If a task author is using the tool lib, and also using a different version of the task lib (different from what the tool lib carries), then we will run into this problem.

Storing the bool on global seems to work. Although I haven't done that before, so I wanted to get an extra set of eyes to validate that seems like a decent solution. I'm hoping that's a decent way to solve the problem in memory - i want to avoid writing a file to disk to solve this (messy, esp for task authors doing manual or L0 tests).

yaananth: This looks okay to me

ericsciple: discussed w/ ting

PR #233, opened 2017-04-04
Make exec and execSync functions take Partial<IExec(Sync)Options>

ILMTitan:


damccorm:
I believe Partial is only available in TypeScript 2, this project uses Typescript 1, so I don't think this update will work. I'm going to close here - feel free to reopen if appropriate.

bryanmacfarlane:
We should update to TypeScript 2

damccorm:
@ILMTitan we are now using TypeScript 3 - are you still interested in reviving this? What's your use case where you needed it?

ILMTitan:
I am no longer working on the project that prompted this.

PR #238, opened 2017-04-14
Add tsc to the PATH

ericsciple:


ericsciple:
this enables running `node make.js build` rather than force running via `npm run build` (consistency with other repos)

ericsciple:
...npm adds node_modules/.bin to the PATH. But when running directly `node.js make <cmd>` it's not added to the path. This change will make sure the PATH is altered.

damccorm:
It looks like this is no longer an issue (node make.js build now works), so I'm going to close. Feel free to reopen if you disagree.

PR #249, opened 2017-06-01
handle tilde expansion for OS home directory

cicorias:
this is related to https://github.com/Microsoft/vsts-tasks/issues/4302

This handles tilde `~` expansion for bash shells where the `PATH` has a tilde such as `~/bin`.  

Since node's `fs.statSync` doesn't expand tilde's, See this conversation as well as to why node team didn't fix. https://github.com/nodejs/node/issues/684#issuecomment-72344577

### DIFF
--- a/node/internal.ts
+++ b/node/internal.ts
@@ -6,6 +6,7 @@ import util = require('util');
 import tcm = require('./taskcommand');
 import vm = require('./vault');
 import semver = require('semver');
+const home = os.homedir();
 
 /**
  * Hash table of known variable info. The formatted env var name is the lookup key.
@@ -401,7 +402,11 @@ export function _which(tool: string, check?: boolean): string {
         }
 
         // return the first match
-        for (let directory of directories) {
+        for (var directory of directories) {
+            // TODO: fix to make tests pass.
+            // if ( !!home && process.platform == 'win32') {
+            //     directory = _unTildify(directory);
+            // }
             let filePath = _tryGetExecutablePath(directory + path.sep + tool, extensions);
             if (filePath) {
                 _debug(`found: '${filePath}'`);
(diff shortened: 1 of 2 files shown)


ericsciple:
I tried this from a shell script (one script calls another script) and it resolves ~ in the paths. So I think adding support for this probably makes sense. At first I was confused at the motivation - I originally assumed it was primarily only supported interactively by the bash shell.

node/internal.ts
```diff
@@ -426,7 +427,7 @@ export function _which(tool: string, check?: boolean): string {
 function _tryGetExecutablePath(filePath: string, extensions: string[]): string {
     try {
         // test file exists
-        let stats: fs.Stats = fs.statSync(filePath);
+        let stats: fs.Stats = fs.statSync(_unTildify(filePath));
```
ericsciple: Since untildify is only run on that path that is stat'd and not the path that is returned, wouldn't which ultimately return the "wrong" tildified path?

cicorias: filePath in tryGetExe is the segment of the PATH env variable which is what would be prefixed with a tilde.

It's fs.statSync that fails if there is a tilde. See the Node issue where they won't fix

So untildify just expands tilde only when it's a prefix as that should only be the case

All the Tests pass which include tests for running ls, grep, etc

ericsciple: If someone calls `tl.which('sometool')` and it returns a tildified-file-path, won't that cause problem depending on what the path is then used for? For instance, if they stat or lstat it will blow up. I would imagine other things may error as well.

I think it would make more sense to do this in the which function when iterating over the directories. And probably only when home is truthy and process.platform != 'win32' and the dir is `~` or starts with `~/`. Also I noticed the regex below also handles `~\` which doesn't seem correct to me.

Also I'm curious if you have insight into how common this problem is? The stance node took is interesting, and I think makes sense for node. However, since the task lib exposes a which() function, it makes me feel like we have a better argument to add it (pragmatic reasoning to avoid common confusion).

cicorias: let me add some tests that deal with these conditions and take it from there. The regex was taken directly from that library referenced by many in the Node issue.

Please stand by....

cicorias: also, within the past week 2 other people using the AZ CLI ran into this problem on macOS. It's fairly routine in `.bash_profile` to rely on `bash` support of `~` instead of something like `$HOME`

cicorias: This is going to take longer as ultimately it's `let stats: fs.Stats = fs.statSync(filePath);` that is failing, yet in each block the `ENOENT` exception is being eaten, so it passes. Some of these might valid; that's what I have to investigate.  e.g.. https://github.com/Microsoft/vsts-task-lib/blob/master/node/internal.ts#L453 and https://github.com/Microsoft/vsts-task-lib/blob/master/node/internal.ts#L495

ericsciple: thats why i dont want which() to return the tildified path :). the dir should be untildified before it's joined with the file name to construct the candidate full path.

damccorm:
I'm going to close this since there hasn't been progress for a year and a half, please reopen if appropriate. If people are still experiencing issues from this, lets log an issue and move forward from there. @cicorias if you're planning on continuing work on this, feel free to reopen here or start a new PR.

PR #252, opened 2017-06-21
add function to import additional loc files

bishal-pdMSFT:


node/internal.ts
```diff
@@ -140,7 +140,7 @@ function _loadLocStrings(resourceFile: string, culture: string): { [key: string]
 /**
  * Sets the location of the resources json.  This is typically the task.json file.
  * Call once at the beginning of the script before any calls to loc.
- * 
+ *
```
ericsciple: can you revert all of the whitespace changes?

node/internal.ts
```diff
@@ -171,8 +171,27 @@ export function _setResourcePath(path: string): void {
 }
 
 /**
+ * Imports localized strings from json file.
+ * It can be used by common modules to load module specific loc strings
+ *
+ * @param     path      Full path to the json.
+ * @returns   void
+ */
+export function _importLocStrings(path: string): void {
```
ericsciple: I haven’t compared the code side by side, but I think this should follow the pattern of the ps3 lib and overlay the values. I believe the ps3 lib also has unit tests for its loc functions.

However, it should follow the pattern of the ps3 lib *in spirit*, not verbatim. The node lib already has a function to import resource strings, we shouldn't add a second one. We should relax the constraint on the existing one.

bishal-pdMSFT: It _is_ overlaying the values in _locStringCache dictionary. I will add unit tests once we close on the approach.
Relaxing constraint in existing function _setResourcePath_ is also an option but will lead to confusion as we will really don't want to set more than one resource file. What we want is to import strings from more than one resx file. IMO, a better approach will be to add one more parameter called 'additionalResourceFiles' in setResourcePath function. Task will call it providing path to its task.json as well as array of all additional module.json files it has referenced from common folder
`tl.setResourcePath("path.join(__dirname, "task.json")", [path.join(__dirname, "node_modules", "azure-arm-rest", "module.json"), path.join(__dirname, "node_modules", "webdeployment-common", "module.json")])`

bishal-pdMSFT: @ericsciple Any thought about above?

ericsciple: Sorry this fell off my radar.

While it is true that the name setResourcePath is not the best terminology, we can't rename the function in a patch-update and I think two functions would be more confusing. We can add intellisense comments to the function.

It's atypical for a task to import multiple resource files. The problem we are actually running into is that a task imports a resource file, and then a module that it references imports a separate resource file. From the task author perspective, they are only importing one resource file. From the module author perspective, they also are only importing one resource file. It's important in the function intellisense comments to indicate it overlays and that authors of re-usable modules should prefix their resource keys to avoid collisions (for instance all task lib resource keys are prefixed `LIB_`)

ericsciple: @bryanmacfarlane thoughts?

bishal-pdMSFT: Another thing which we should keep in mind for this approach is that task can directly load a file inside module rather than loading the complete module. i.e. require("nuget-task-common/Authentication") instead of require("nuget-task-common"). Infact this is the way most of the tasks are importing modules currently. What this means is that module author will have to call setResourcePath in each of the files in module.

bishal-pdMSFT: @ericsciple Are you fine with the idea of potentially each file in common module in calling setResourcePath()? If so, I will relax constraint in setResourcePath to allow more than one resource file.

ericsciple: Yes. It's up to the module author. Even without a single predictable entry file for the module, each possible entry file can reference a common file that imports the resource strings. If two separate files require the same common file, node caches the require and it's only actually loaded once.

The other suggestion that was previously brought up was at build time in the vsts-tasks repo to inject lib.json messages into task.json. The problem I have with that approach is that we need to find a solution that works for customers too.

If we can find out the __dirname of the caller then we can automatically look for and load a task.json/lib.json without requiring anyone to call setResourcePath. I don't know whether that is feasible.

bryanmacfarlane: We shouldn't add a function.  If anything we should move to the model of a function call only needed in the exceptional case.  In the common and virtually all cases that I know of, a task.json or module.json is in the current dir.

Is that feasible?

ericsciple: If I call tl.loc(), then tl.loc doesn't know what directory I am in (the caller's directory).

bryanmacfarlane: I agree with this.

" It's important in the function intellisense comments to indicate it overlays and that authors of re-usable modules should prefix their resource keys to avoid collisions (for instance all task lib resource keys are prefixed LIB_)"

bryanmacfarlane: also note that re-usable modules as we have implemented in this repo is our concept (not a general node concept which relies on npm packages).

bishal-pdMSFT: how about overload tl.loc with an optional parameter called callerDir. caller will call tl.loc("loc-string",  __dirname) ?

ericsciple: interferes with the format args parameters

bryanmacfarlane: why not just prefix?

ericsciple: `tl.loc(__dirname, 'resourceKey', formatArg1, formatArg2)` breaks compat. The current interface expects the resource key to be first and there is no normal way to distinguish whether you really meant __dirname instead of resourceKey.

However, there is an unorthodox way to distinguish. If the first arg is a cache miss, then check if it's a directory...

bishal-pdMSFT: Or we could introduce a new function like loc2. `tl.loc2(__dirname, 'resourceKey', formatArg1, formatArg2)` as it basically is an enhanced functionality. Discoverability will be tricky, but can be mitigated using intellisense.

bryanmacfarlane: Does prefix with LIB_ work?

ericsciple: The [resource keys](https://github.com/Microsoft/vsts-task-lib/blob/master/node/lib.json#L3) themselves are prefixed with LIB_ as a means to avoid collisions with user resource keys.

ericsciple: Discussed offline with Bryan and went over the current design. Let's move forward with relaxing the constraint so setResourcePath can be called on more than one file.

bishal-pdMSFT: Okay, I will proceed with it. May I know what were your thoughts against loc2 approach. I know that adding such functions for purpose of compat is not good. But given that we are dealing with a function which takes variable number of arguments, it seemed okay to me. thoughts?

I might not take it up in next couple of days and I am stuck with a high priority work. Hope that is fine.

PR #253, opened 2017-06-21
Adding error code to distinguish between errors

nigurr:


nigurr:
* Will discuss with @ericsciple  whether to use "Complete Task command result code or use logissue result code"

ericsciple:
Can you only make the minimum change and not reformat the document? I'm having a hard time figuring out what the actual change is.

nigurr:
@ericsciple sorry. couldn't resist pressing formatting shortcut. 

Changes: 
https://github.com/Microsoft/vsts-task-lib/pull/253/files#diff-f7bc92bcb6131b3321462f4336b98b7bR281
https://github.com/Microsoft/vsts-task-lib/pull/253/files#diff-8d040e410d00a119c5517a408d3a7ec0R41

Exposing `resultcode` for warning/error as `logissue` already supports it.
Adding `resultcode` for complete command. this shouldn't break compat as it's new property

nigurr:
@ericsciple can you approve this? :)

ericsciple:
@bryanmacfarlane @TingluoHuang to review since i'm not involved in the current proposed telemetry direction, and this is different than the initial discussion of a ## command

ericsciple:
also does this require a new agent version? you may need to check agent version before including code in the command. would need to be documented on the function comments.

nigurr:
It doesn't require new agent as it's additional property and will not be parsed in older agents.

bryanmacfarlane:
I don't want to hijack warning and error.   It's not a warning or error.

Whether it's not parsed in older agents or not called (##vso) because you don't have a newer agent, the result is the same - older agents won't send telemetry.

Use a new ##vso command

nigurr:
@bryanmacfarlane addressed the review comments. it uses only complete command error code now.

nigurr:
@bryanmacfarlane can you please approve the pr?

nigurr:
@TingluoHuang Can you please merge the PR?

PR #258, opened 2017-06-24
Add PatternRoot to Select-Match

ericsciple:


ericsciple:
Ported from node lib. Tests ported as well.

TingluoHuang:
Looks good.

PR #261, opened 2017-06-27
Adding telemetry publish command to task lib

nigurr:


nigurr:
@ericsciple, @TingluoHuang  can you please review and approve the same?

node/task.ts
```diff
@@ -1645,3 +1645,26 @@ if (semver.lt(process.versions.node, '4.2.0')) {
...
+ */
+export function publishTelemetry(area: string, feature: string, properties: { [key: string]: any }): void {
+
+    let agent = getVariable('Agent.Version');
+    if (!agent || semver.lt(agent, '2.120.0')) {
+        return;
+    }
+
+    let data = JSON.stringify(properties);
+    debug('telemetry area: ' + area + ' feature: ' + feature + ' data: ' + data);
```
ericsciple: @TingluoHuang to confirm, i think the pattern we normally use is the agent prints the debug info related to the `##` commands.

TingluoHuang: yes, agent will print out any ## commands it run when system.debug=true

node/test/telemetrytests.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    it('publish telemetry if the agent version is greater than 2.120', function (done) {
+        this.timeout(1000);
+
+        let originalAgentVersion = process.env['AGENT_VERSION'];
+        process.env['AGENT_VERSION'] = '2.120.0';
+
+        try {
+            let data = { 'abc': 'xyz' };
+            var stdStream = testutil.createStringStream();
```
ericsciple: does this need to unset stdstream in the finally?

node/test/telemetrytests.ts
```diff
@@ -0,0 +1,65 @@
+/// <reference path="../typings/index.d.ts" />
+/// <reference path="../_build/task.d.ts" />
+
+import assert = require('assert');
+import * as tl from '../_build/task';
+import testutil = require('./testutil');
+
+describe('Telemetry Tests', function () {
```
ericsciple: can you move this into commandtests.ts to keep the number of files down (there are only two tests in this file)

ericsciple:
@TingluoHuang to review. afaik i know this makes sense but i believe ting was involved more on the `##` command on the agent side.

node/test/telemetrytests.ts
nigurr: commandtest tests are targeted at 'taskcommand' file rather than specific command. so i kept separate file.

nigurr:
@TingluoHuang Can you please complete the review/PR?

PR #268, opened 2017-07-14
add proxy support

TingluoHuang:


node/task.ts
```diff
@@ -1531,6 +1531,41 @@ export function findMatch(defaultRoot: string, patterns: string[] | string, find
 }
 
 //-----------------------------------------------------
+// Http Proxy Helper
+//-----------------------------------------------------
+
+export interface ProxyConfiguration {
+    proxyUrl: string;
+    proxyUsername?: string;
```
ericsciple: should this be `proxyUserName` since user name are two words? or should this interface match some other convention (e.g. match an interface exposed by node?)

node/task.ts
```diff
@@ -1531,6 +1531,41 @@ export function findMatch(defaultRoot: string, patterns: string[] | string, find
...
+}
+
+export function getHttpProxyConfiguration(): ProxyConfiguration {
+    let proxyUrl: string = getVariable('Agent.ProxyUrl');
+    if (proxyUrl && proxyUrl.length > 0) {
+        let proxyUsername: string = getVariable('Agent.ProxyUsername');
+        let proxyPassword: string = getVariable('Agent.ProxyPassword');
+
+        let proxyBypassHosts: string[]
+        let proxyBypassHostsJson = getVariable('Agent.ProxyBypassList');
```
ericsciple: curious if you considered a not-null pattern here, e.g.
```typescript
let proxyBypassHosts: string[] = JSON.parse(getVariable('Agent.ProxyBypassList') || '[]');
```

powershell/CompiledHelpers/VstsTaskSdk.cs
```diff
@@ -1,12 +1,94 @@
...
 namespace VstsTaskSdk
 {
     public class TerminationException : Exception
     {
         public TerminationException(String message) : base(message) { }
     }
+
+    public sealed class VstsWebProxy : IWebProxy
+    {
+        private string _proxyAddress;
```
ericsciple: _proxyAddress is effectively readonly too, correct?

powershell/VstsTaskSdk/ServerOMFunctions.ps1
```diff
@@ -421,6 +426,26 @@ function Get-VssHttpClient {
...
 }
 
+function Get-WebProxy {
+    [CmdletBinding()]
+    param()
+
+    Trace-EnteringInvocation -InvocationInfo $MyInvocation
+    try
+    {
+        $ProxyUrl = Get-TaskVariable -Name Agent.ProxyUrl
```
ericsciple: camel case local variables

powershell/VstsTaskSdk/ServerOMFunctions.ps1
```diff
@@ -421,6 +426,26 @@ function Get-VssHttpClient {
...
+        $ProxyUserName = Get-TaskVariable -Name Agent.ProxyUserName
```
ericsciple: Do you need to declare these as strings so the right constructor is always chosen below, even when these are null or empty?

E.g. `[string]$proxyUserName = Get-TaskVariable -Name Agent.ProxuUserName`

ericsciple: nevermind.... this is the custom class, there is only one ctor.

ericsciple:
minor feedback, use your judgement whether to take. looks good.

bryanmacfarlane:
min agent version for proxy and bypass list in docs

OK

PR #276, opened 2017-07-27
Revert "Adding telemetry publish command to task lib"

nigurr:
Reverts Microsoft/vsts-task-lib#261


nigurr:
Reverting this PR as dicussed with @cltshivash

PR #279, opened 2017-08-18
Add escaping of closing bracket character

stephenmichaelf:
This investigation originally started because of [this](https://github.com/Microsoft/vsts-agent/issues/1100) issue.

The code here doesn't solve that specific issue but we need it anyways. That issue will be solved by a change in DistributedTask.

This change adds escaping for ']' in the node code and ps code. We also weren't escaping the property values in the node code previously so that is added too. The Agent already unescapes property values.

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vsts-task-lib",
-  "version": "2.0.7",
+  "version": "2.0.8",
   "description": "VSTS Task SDK",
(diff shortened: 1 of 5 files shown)


stephenmichaelf:
Talked with @TingluoHuang and removed escaping of ; and ] in the body, left only escaping of them in the property value.

TingluoHuang:
Looks good.

PR #308, opened 2018-01-29
remove extra schema info

ericsciple:


tasks.schema.json
```diff
@@ -29,17 +29,6 @@
     "author": {
       "type": "string"
     },
-    "visibility": {
```
AArnott: @ericsciple We have some tasks that define this property today. Should we drop it? Is it completely ignored at this point?

ericsciple: preview is definitely not correct.

I think release ignores the visibility rules, and build does not.

most tasks should run on either, so i think it generally should not be required. we'll probably stop filtering in build, and rely on task authors failing gracefully at runtime.

PR #320, opened 2018-03-02
add trigger into yaml, change queues

ericsciple:


.vsts-ci.yml
```diff
@@ -1,13 +1,13 @@
...
-- phase: build_on_windows
+- phase: windows
 #################################################
-  displayName: (vsts-task-lib) Windows
-  queue:
-    #name: Hosted
-    name: buildDevs
-    demands: agent.os -equals Windows_NT
+  condition: and(succeeded(), eq(variables.os, 'Windows_NT'))
+  queue: Hosted VS2017
```
ericsciple: switched to hosted

.vsts-ci.yml
```diff
@@ -1,13 +1,13 @@
-phases:
+trigger:
+- master
+- features/*
```
ericsciple: trigger info

node/test/scripts/match-input-exe.cs
```diff
@@ -0,0 +1,59 @@
+using System;
```
ericsciple: this is a small program used on windows only for reliably testing the tool runner feature for piping the output from one program to another

this is the second program in the chain

node/test/scripts/print-output-exe.cs
```diff
@@ -0,0 +1,39 @@
+using System;
```
ericsciple: this is a small program used on windows only for reliably testing the tool runner feature for piping the output from one program to another

this is the first program in the chain

.vsts-ci.yml
```diff
@@ -99,10 +99,10 @@ phases:
     workingDirectory: node
 
 #################################################
-- phase: build_on_osx
+- phase: macOS
 #################################################
-  displayName: (vsts-task-lib) osx
-  queue: Hosted Mac Internal
+  condition: and(succeeded(), eq(variables.os, 'Darwin'))
+  queue: Hosted macOS Preview
```
ericsciple: i might need to switch this back to "Hosted Mac Internal". Every public account has "Hosted macOS Preview", but we only have 1 agent in that pool so i'm not sure what is going on. i'll check with mms folks tomorrow.

powershell/Tests/lib/Initialize-Test.ps1
```diff
@@ -14,6 +14,12 @@ Import-Module 'Microsoft.PowerShell.Utility' -Verbose:$false
 Write-Verbose "Importing module: Microsoft.PowerShell.Security"
 Import-Module 'Microsoft.PowerShell.Security' -Verbose:$false
 
+if ($env:AGENT_TEMPDIRECTORY) {
```
ericsciple: this works around the 8.3 filename problem on the hosted agent

powershell/Tests/lib/Start-Engine.ps1
```diff
@@ -57,8 +57,19 @@ function Invoke-Test {
 
 # Record the original environment variables.
 $originalEnv = @{ }
-foreach ($envVar in (Get-ChildItem -LiteralPath env:)) {
-    $originalEnv[$envVar.Name] = $envVar.Value
+foreach ($key in ([Environment]::GetEnvironmentVariables()).Keys) {
```
ericsciple: this was a really bizarre problem. i didnt even know this was possible on windows.

powershell/Tests/lib/Start-Engine.ps1
```diff
@@ -71,23 +82,26 @@ while ($true) {
...
     foreach ($key in $originalEnv.Keys) {
         if (!$currentMatches.ContainsKey($key)) {
-            Set-Content -LiteralPath "env:$key" -Value "$($originalEnv[$key])"
+            [Environment]::SetEnvironmentVariable($key, $originalEnv[$key])
         }
     }
 
     # Write a special "end-of-test" message over STDOUT.
     Write-Host '_END_OF_TEST_ce10a77a_'
+    [Console]::Out.Flush()
```
ericsciple: this is a lesson learned from problems we had in the tasks repo. when the machine is under high load it appears powershell (or .net?) may delay the flush for a long while. this came up when the tasks were timing out in the tasks repo, and i had even bumped the timeouts to 2 minutes and still timing out - finally flush did the trick and i was able to lower the timeouts to something sane

node/test/toolrunnertests.ts
```diff
@@ -395,8 +395,68 @@ describe('Toolrunner Tests', function () {
                 done(err);
             });
     })
+    // function to compile a .NET program that prints lines.
```
ericsciple: we do a similar thing already to test the tool runner passing args correctly, so i reused that pattern here. this compiles a .net app on the fly using csc.

node/test/toolrunnertests.ts
```diff
@@ -33,7 +33,7 @@ describe('Toolrunner Tests', function () {
 
 
     it('ExecSync convenience with stdout', function (done) {
-        this.timeout(1000);
+        this.timeout(10000);
```
ericsciple: slower sometimes

node/test/toolrunnertests.ts
```diff
@@ -395,8 +395,68 @@ describe('Toolrunner Tests', function () {
                 done(err);
             });
     })
+    // function to compile a .NET program that prints lines.
+    // the helper program is used on Windows to validate piping output between tools.
+    let compileOutputExe = (): string => {
```
ericsciple: there is a lot of duplication here. i'll refactor the common code into a separate function.

node/test/toolrunnertests.ts
```diff
@@ -409,21 +469,25 @@ describe('Toolrunner Tests', function () {
...
-            tasklist.pipeExecOutputToTool(find);
+            var matchExe = tl.tool(compileMatchExe())
+                .arg('0') // exit code
+                .arg('line 2'); // match value
+            var outputExe = tl.tool(compileOutputExe())
+                .arg('0') // exit code
+                .arg('line 1')
+                .arg('line 2')
+                .arg('line 3');
+            outputExe.pipeExecOutputToTool(matchExe);
```
ericsciple: the match program i wrote waits for STDIN to close. i think find.exe may have been too hasty for how we were trying to test here. it was previously flaky, and then reliably failing when i switched to hosted. the custom program appears to be reliably working.

node/test/toolrunnertests.ts
```diff
@@ -395,8 +395,68 @@ describe('Toolrunner Tests', function () {
...
+            [
+                '/target:exe',
+                `/out:${exePath}`,
+                sourceFile
+            ]);
+        return exePath;
+    }
     it('Exec pipe output to another tool, succeeds if both tools succeed', function(done) {
-        this.timeout(1000);
+        this.timeout(20000);
```
ericsciple: unfortunately calling csc twice and then invoking two processes can take a while sometimes depending upon the machine

powershell/Tests/lib/Start-Engine.ps1
ericsciple: after i ran into this problem, i switched this script over to use the .net method everywhere for accessing env vars instead of the ps way.

PR #321, opened 2018-03-02
Refactor compile exe helper functions

ericsciple:


ericsciple:
minor cleanup from previous checkin

PR #324, opened 2018-03-06
add optional parameter to getHttpProxyConfig allow bypass filter.

TingluoHuang:
not test yet.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -1563,19 +1563,33 @@ export interface ProxyConfiguration {
  *
  * @return  ProxyConfiguration
  */
-export function getHttpProxyConfiguration(): ProxyConfiguration {
+export function getHttpProxyConfiguration(requestUrl?: string): ProxyConfiguration {
     let proxyUrl: string = getVariable('Agent.ProxyUrl');
     if (proxyUrl && proxyUrl.length > 0) {
         let proxyUsername: string = getVariable('Agent.ProxyUsername');
         let proxyPassword: string = getVariable('Agent.ProxyPassword');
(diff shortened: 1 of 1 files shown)


node/task.ts
```diff
@@ -1563,19 +1563,33 @@ export interface ProxyConfiguration {
...
 
-        return {
-            proxyUrl: proxyUrl,
-            proxyUsername: proxyUsername,
-            proxyPassword: proxyPassword,
-            proxyBypassHosts: proxyBypassHosts
-        };
+        let bypass: boolean = false;
+        if (requestUrl) {
+            this._httpProxyBypassHosts.forEach(bypassHost => {
```
stephenmichaelf: Didn't see this in the code? How is this different than proxyBypassHosts?

jessehouwing: Does that include wildcards, ip ranges etc as well?

bryanmacfarlane: it's a list of regex patterns

node/task.ts
```diff
@@ -1563,19 +1563,33 @@ export interface ProxyConfiguration {
  *
  * @return  ProxyConfiguration
  */
-export function getHttpProxyConfiguration(): ProxyConfiguration {
+export function getHttpProxyConfiguration(requestUrl?: string): ProxyConfiguration {
```
keithrob: Should this be URL?

keithrob: OK, You could do requestUrl.host if it was a URL() though.

keithrob:
@TingluoHuang,
What are the merge plans for this PR?

bryanmacfarlane:
Plan is to click merge which I just did.

PR #329, opened 2018-03-14
Add option to pipe to file and another tool

madhurig:
We want to be able to write the output of the first tool to a file in addition to piping it to another tool to simulate something like this: xcodebuild <options> | tee xcodebuild.log | xcpretty. This works but looking for feedback on the approach before adding unit tests.

### DIFF
--- a/node/toolrunner.ts
+++ b/node/toolrunner.ts
@@ -178,6 +178,29 @@ export class ToolRunner extends events.EventEmitter {
         return cmd;
     }
 
+    private _processLineBuffer(data: Buffer, strBuffer: string, onLine: (line: string) => void): void {
+        try {
+            var s = strBuffer + data.toString();
+            var n = s.indexOf(os.EOL);
+
+            while (n > -1) {
+                var line = s.substring(0, n);
+                onLine(line);
+
+                // the rest of the string ...
+                s = s.substring(n + os.EOL.length);
+                n = s.indexOf(os.EOL);
+            }
+
+            strBuffer = s;
+        }
+        catch (err) {
+            // streaming lines to console is best effort.  Don't fail a build.
+            this._debug('error processing line');
+        }
+
+    }
+
     private _getSpawnFileName(): string {
         if (process.platform == 'win32') {
             if (this._isCmdFile()) {
@@ -571,6 +594,9 @@ export class ToolRunner extends events.EventEmitter {
         let successFirst = true;
         let returnCodeFirst: number;
         let fileStream: fs.WriteStream;
+        let waitingEvents: number = 0; // number of process or stream events we are waiting on to complete
+        let returnCode: number = 0;
+        let error;
 
         if (this.pipeOutputToTool) {
             toolPath = this.pipeOutputToTool.toolPath;
@@ -580,10 +606,13 @@ export class ToolRunner extends events.EventEmitter {
             // https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options
 
             //start the child process for both tools
+            waitingEvents++;
             var cpFirst = child.spawn(
                 this._getSpawnFileName(),
                 this._getSpawnArgs(options),
                 this._getSpawnOptions(options));
+            
+            waitingEvents ++;
             cp = child.spawn(
                 this.pipeOutputToTool._getSpawnFileName(),
                 this.pipeOutputToTool._getSpawnArgs(options),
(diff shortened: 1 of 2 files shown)


node/toolrunner.ts
```diff
@@ -584,27 +588,48 @@ export class ToolRunner extends events.EventEmitter {
                 this.pipeOutputToTool._getSpawnArgs(options),
                 this.pipeOutputToTool._getSpawnOptions(options));
 
+            let fileStream: fs.WriteStream = this.pipeOutputToFile ? fs.createWriteStream(this.pipeOutputToFile) : null;
```
ericsciple: will this work if the file already exists? It should append in that case I guess?

node/toolrunner.ts
```diff
@@ -584,27 +588,48 @@ export class ToolRunner extends events.EventEmitter {
...
+            if (fileStream) {
+                fileStream.write(this._getCommandString(options) + os.EOL);
```
ericsciple: is it weird to pipe the command string? I would expect the file to only contain the output of the first tool.

node/toolrunner.ts
```diff
@@ -584,27 +588,48 @@ export class ToolRunner extends events.EventEmitter {
...
+            }
+
             //pipe stdout of first tool to stdin of second tool
             cpFirst.stdout.on('data', (data: Buffer) => {
                 try {
+                    if (fileStream) {
+                        fileStream.write(data.toString());
```
ericsciple: why `.toString()`?

node/toolrunner.ts
```diff
@@ -584,27 +588,48 @@ export class ToolRunner extends events.EventEmitter {
...
+                    }
                     cp.stdin.write(data);
                 } catch (err) {
                     this._debug('Failed to pipe output of ' + toolPathFirst + ' to ' + toolPath);
                     this._debug(toolPath + ' might have exited due to errors prematurely. Verify the arguments passed are valid.');
                 }
             });
             cpFirst.stderr.on('data', (data: Buffer) => {
+                if (fileStream) {
+                    fileStream.write(data.toString());
```
ericsciple: same question about `.toString()`

node/toolrunner.ts
```diff
@@ -584,27 +588,48 @@ export class ToolRunner extends events.EventEmitter {
...
+                }
                 successFirst = !options.failOnStdErr;
                 if (!options.silent) {
                     var s = options.failOnStdErr ? options.errStream : options.outStream;
                     s.write(data);
                 }
             });
             cpFirst.on('error', (err) => {
+                if (fileStream) {
+                    fileStream.on('finish', () => {
```
ericsciple: copy paste error? shouldn't this write stderr to the file too?

ericsciple: oops I misread this.

node/toolrunner.ts
```diff
@@ -584,27 +588,48 @@ export class ToolRunner extends events.EventEmitter {
...
+                if (fileStream) {
+                    fileStream.on('finish', () => {  
+                        fileStream.close();
+                    });
+                }
                 cp.stdin.end();
                 defer.reject(new Error(toolPathFirst + ' failed. ' + err.message));
             });
             cpFirst.on('close', (code, signal) => {
+                if (fileStream) {
```
ericsciple: I wonder if this is reliable. Do you know whether `close` means no further stdout/stderr data events are pending to be drained? I believe the node docs have good details about this, but I don't remember the details offhand.

node/toolrunner.ts
madhurig: The command string helps set context for what this output was for. Otherwise user has to use the step number to match the output especially if there are multiple tasks of the same type that are producing these files. I initially did not write it but after I tried e2e it made more sense to write it.

node/toolrunner.ts
madhurig: that's right, it should not be needed. I will pull it out and verify.

node/toolrunner.ts
madhurig: It should be as per https://nodejs.org/api/child_process.html#child_process_event_close - The 'close' event is emitted when the stdio streams of a child process have been closed. This is distinct from the 'exit' event, since multiple processes might share the same stdio streams.

node/toolrunner.ts
ericsciple: we cant do this. it may contain secrets and does not get run through the log scrubber

madhurig: Ah good point!

node/toolrunner.ts
madhurig: I wrote tests and found that there were some timing issues. I fixed the issues but the code looks ugly. Will sync up before doing any refactoring.

node/toolrunner.ts
```diff
@@ -82,10 +84,10 @@ export class ToolRunner extends events.EventEmitter {
         var args = [];
 
         var inQuotes = false;
-        var escaped =false;
+        var escaped = false;
```
madhurig: @ericsciple: I ran format command in VSCode that brought in other changes. If you don't want these committed, I can back them out .

PR #331, opened 2018-03-15
Users/stfrance/oss

stephenmichaelf:


stephenmichaelf:
@bryanmacfarlane  It looks like vsts-tasks does copy node_modules to the outDir manually:

image

I deleted my _build folder for typed-rest-client and realized it also isn't copied there. I wonder if there was a code change recently that removed the copying of node_modules. It's strange that it's happening for both repos though. Since there was an _build folder in outDir for both vsts-task-lib and typed-rest-client it must mean at some point it was being copied. Will keep digging.

PR #332, opened 2018-03-16
Update TaskLibAnswers to have all supported commands

brcrista:
These are all the commands that are looked up in `MockAnswers` in mock-task and mock-toolrunner.
With this change, we should stop using type assertions to force the `TaskLibAnswers` type when we set up the mocks for our L0 tests.
This will allow test authors to catch commands they think are being mocked but are not.


brcrista:
Someone will have to merge this once it has sign off.  I don't have write permissions for this repo.

stephenmichaelf:
LGTM

PR #334, opened 2018-03-26
Missed a couple mocked commands.  Bump patch version

brcrista:
Forgot to bump the patch version before.
Noticed that I missed a couple of commands that were actually mocked, and so the previous changes were insufficient to stop coercing the `TaskLibAnswer` type in our existing code.  I defined the `MockedCommand` type here so that the compiler will catch all arguments to `getResponse`.

### DIFF
--- a/node/mock-answer.ts
+++ b/node/mock-answer.ts
@@ -22,7 +22,7 @@ export interface TaskLibAnswers {
 }
 
 // TODO TypeScript 2.1: use `keyof`
-export type MockedCommand = "checkPath"
+export type MockedCommand = 'checkPath'
     | 'cwd'
     | 'exec'
     | 'exist'
@@ -51,7 +51,7 @@ export class MockAnswers {
         }
 
         if (!this._answers[cmd]) {
-            debug(`no mock responses registered for given cmd`);
+            debug(`no mock responses registered for ${JSON.stringify(cmd)}`);
             return null;
         }
 
--- a/node/mock-answer.ts
+++ b/node/mock-answer.ts
@@ -21,7 +21,7 @@ export interface TaskLibAnswers {
     which?: { [key: string]: string },
 }
 
-// TODO TypeScript 2.1: use `keyof`
+// TODO TypeScript 2.1: replace with `keyof TaskLibAnswers`
 export type MockedCommand = 'checkPath'
     | 'cwd'
     | 'exec'
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vsts-task-lib",
-  "version": "2.3.1",
+  "version": "2.3.2",
   "description": "VSTS Task SDK",
   "main": "./task.js",
   "typings": "./task.d.ts",
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vsts-task-lib",
-  "version": "2.3.2",
+  "version": "2.4.0",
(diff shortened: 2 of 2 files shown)


node/mock-answer.ts
```diff
@@ -1,5 +1,5 @@
-import path = require('path');
-import fs = require('fs');
+import * as path from 'path';
```
brcrista: `import ... require` has slightly different semantics than `import ... from`.  My understanding is that `require` executes the module as the CommonJS and AMD `require` function does so that any side effects from the module are applied. `import ... from` is the normal way to import a module in TypeScript.

node/mock-answer.ts
```diff
@@ -15,10 +15,25 @@ export interface TaskLibAnswers {
...
     findMatch?: { [key: string]: string[] },
     ls?: { [key: string]: string },
+    osType?: { [key: string]: string },
     rmRF?: { [key: string]: { success: boolean } },
-    which?: { [key: string]: string; },
+    stats?: { [key: string]: any }, // Can't use `fs.Stats` as most existing uses don't mock all required properties
+    which?: { [key: string]: string },
 }
 
+// TODO TypeScript 2.1: use `keyof`
```
stephenmichaelf: What does this mean?

node/mock-answer.ts
```diff
@@ -15,10 +15,25 @@ export interface TaskLibAnswers {
...
+export type MockedCommand = "checkPath"
```
stephenmichaelf: Double quotes here and single below?

node/mock-answer.ts
```diff
@@ -29,7 +44,7 @@ export class MockAnswers {
         this._answers = answers;
     }
 
-    public getResponse(cmd: string, key: string, debug: (message: string) => void): any {
+    public getResponse(cmd: MockedCommand, key: string, debug: (message: string) => void): any {
```
stephenmichaelf: Breaking change? May want to bump at least the minor since it's mock and not the main code.

stephenmichaelf: Maybe MockCommand?

node/mock-answer.ts
brcrista: I believe we're using version 1.8.7 of the TypeScript compiler.  TypeScript 2.1 adds a `keyof` operator: you can just say `keyof TaskLibAnswer` instead of defining `MockedCommand`.

node/mock-answer.ts
brcrista: thanks, will fix

node/mock-answer.ts
brcrista: Yeah, it's technically a breaking change if people are calling `getResponse` directly for some reason.  There aren't any uses right now in vsts-tasks -- it's just used internally in mock-task and mock-toolrunner.

I said `MockedCommand` since it's "one of the commands that are mocked" and not the mock itself.

madhurig: There might be 3rd party extensions using it

brcrista: Yeah, we can bump the minor version

stephenmichaelf:
LGTM.

One last thing:

@bryanmacfarlane Can you confirm that bumping minor version is sufficient when making a breaking change? Don't mean to be pedantic but want to make sure we are staying consistent with the versioning for this lib. Especially given what @madhurig helpfully mentioned about 3rd party extensions.

bryanmacfarlane:
fine.

Issue #317, opened 2018-02-26
findMatch API throws exception, if the pattern matches the broken symlink file


navin22:
### Repro Steps

Call findMatch API with the pattern that matches the broken symlink file. 

#### Actual : 
As the follow symlinks option is true by default, it tries to find and it then end up throwing exception to the caller. 

#### Expected : 
The API should return the matching files at least. Without throwing the exception.  
And conveying the caller the error message, by callback or some other mechanism. 
Or we can have one more API tryFindMatch which does handle these things without throwing the error.

stephenmichaelf:
@navin22 Can you give a more detailed example to reproduce? Including method params and folder/file structure?

Also, follow sym links defaults to true but you can override to false.

navin22:
@stephenmichaelf We call this API from `PublishTestResults` task. Where we use this to get matching TRX or XML test result files.

1. Create a symlink for XML file in folder A in folder B. 
2. Then delete the file in folder A.
3. then call the `findMatch` API with pattern `B\*.xml`. This will throw an exception.

We want to follow the symlink paths. Hence I am not passing `FindOptions` with false. But if there are some other files matching the pattern in folder B and exists. I want to use those files and give a warning for the files that are missing or broken symlinks.

navin22:
@stephenmichaelf Any update on this?

bryanmacfarlane:
@navin22 - we accept PRs if you want to try and fix ahead of @stephenmichaelf getting to this.  He's on some other higher priority release and OSS tasks.

bryanmacfarlane:
The api is behaving correctly here.  If you want another option, create a PR.  It shouldn't warn as the task author needs to decide whether the option.  Don't pass the warning to the consumer of the task

PR #338, opened 2018-04-05
Adding one more API to find matches which ignores statSync erros.

navin22:
It is my attempt to solve the issue #317. Please suggest any design feedback or any other methods for solving it.

- Added new API to task lib tryFindMatch which ignores errors during while following the symlinks.
- Returns the list of matches found and the list of errors if any.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -740,8 +740,6 @@ export interface FindOptions {
  * @returns   string[]
  */
 export function find(findPath: string, options?: FindOptions): string[] {
-    options = options || _getDefaultFindOptions();
-    _debugFindOptions(options)
     return _internalFind(findPath, options, false).matchesFound;
 }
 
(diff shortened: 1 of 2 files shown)


node/task.ts
```diff
@@ -740,104 +740,7 @@ export interface FindOptions {
...
-            }
-        }
-
-        debug(`${result.length} results`);
-        return result;
-    }
-    catch (err) {
-        throw new Error(loc('LIB_OperationFailed', 'find', err.message));
-    }
+    return _internalFind(findPath, options, false).matchesFound;
```
ericsciple: we already have a pattern for find options

ericsciple: follow the existing pattern

damccorm:
Looks like this is going to be a no-op, so I'm going to close here, feel free to reopen if appropriate.

PR #339, opened 2018-04-06
Add `getPlatform` function

brcrista:
Right now, the way we check the platform in tasks is to use `taskLib.osType.match(/^Win/)`.  This is more complicated than it needs to be and it isn't clear how to tell macOS vs. Linux.

(This PR has a bunch of extra commits that have already been merged.  I think it is because they existed individually on my fork but were squashed into a single commit upstream.  Since we will squash again merging upstream it shouldn't matter, but I will fix my fork so it stops happening for future PRs.)

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -522,6 +522,7 @@ export function writeFile(file: string, data: string | Buffer, options?: string 
 }
 
 /**
+ * @deprecated Use `getPlatform`
  * Useful for determining the host operating system.
  * see [os.type](https://nodejs.org/api/os.html#os_os_type)
  * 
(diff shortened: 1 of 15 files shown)


node/task.ts
```diff
@@ -531,6 +532,25 @@ export function osType(): string {
...
+    Linux
+}
+
+/**
+ * Determine the operating system the build agent is running on.
+ * @returns Platform, or throws an `Error` if the platform is not recognized
+ */
+export function getPlatform(): Platform {
+    switch (process.platform) {
+        case 'win32': return Platform.Windows;
```
jpricketMSFT: According to the documentation (https://nodejs.org/api/process.html#process_process_platform) there are quite a few more strings that could be returned. Honestly, I don't know if we support some of them or not. I.e. AIX and SunOS might work but be unsupported. Perhaps the error message should indicate that the platform is not supported instead of not recognized.

brcrista: sure

AlexRukhlin:
LGTM

brcrista:
Left this for too long, going to close and restart.

PR #344, opened 2018-04-16
Enclose symlink statSync with try-catch

TheVorpalMethod:
Re: "Failed find: ENOENT: no such file or directory, stat" when task encounters a broken symlink.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -784,17 +784,27 @@ export function find(findPath: string, options?: FindOptions): string[] {
             // stat returns info about the target of a symlink (or symlink chain),
             // lstat returns info about a symlink itself
             let stats: fs.Stats;
-            if (options.followSymbolicLinks) {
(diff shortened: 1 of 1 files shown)


node/task.ts
```diff
@@ -784,17 +784,27 @@ export function find(findPath: string, options?: FindOptions): string[] {
...
+                    stats = fs.lstatSync(item.path);
+                }
             }
-            else {
-                // use lstat (not following symlinks)
-                stats = fs.lstatSync(item.path);
+            catch (err) {
+                if (err.code == 'ENOENT') {
+                    debug('0 results (potential symlink failure)')
+                    return [];
```
ericsciple: this will not return the other result.

ericsciple: I have a different fix in mind. i'll raise a pr

PR #354, opened 2018-05-09
Link to author agentless tasks

shashban:


damccorm:
I don't think that it makes sense to link to this from the main README, its not really one of the main points of this library. The current documentation seems to be more appropriate.

PR #365, opened 2018-05-22
Find-Match in powershell does not support searching a pattern in a path that does not exist

Thilas:
The fix I propose here intends to get the same behavior than when using ts `findMatch` function. To illustrate that, I added a unit test in both ts and ps.

More precisely, I pinpoint the difference between ts and ps here: https://github.com/Microsoft/vsts-task-lib/blob/master/node/task.ts#L751.

### DIFF
--- a/powershell/VstsTaskSdk/FindFunctions.ps1
+++ b/powershell/VstsTaskSdk/FindFunctions.ps1
@@ -155,7 +155,7 @@ function Find-Match {
                             $findResults += $findPath
                         }
                     } else {
-                        $findResults = Get-FindResult -Path $findPath -Options $FindOptions
+                        $findResults = @( Get-FindResult -Path $findPath -Options $FindOptions )
                     }
 
                     Write-Verbose "Found $($findResults.Count) paths."
@@ -625,7 +625,7 @@ function Get-FindResult {
 
     if (!(Test-Path -LiteralPath $Path)) {
         Write-Verbose 'Path not found.'
-        return ,@( )
+        return
     }
 
     $Path = ConvertTo-NormalizedSeparators -Path $Path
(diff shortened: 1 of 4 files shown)


Thilas:
Any news about this?

Thilas:
@ericsciple, any news about this?

powershell/VstsTaskSdk/FindFunctions.ps1
```diff
@@ -625,7 +625,7 @@ function Get-FindResult {
 
     if (!(Test-Path -LiteralPath $Path)) {
         Write-Verbose 'Path not found.'
-        return
+        return ,@( )
```
ericsciple: what about instead changing the line:
`$findResults = Get-FindResult -Path $findPath -Options $FindOptions`

to:
`$findResults = @( Get-FindResult -Path $findPath -Options $FindOptions )`

I see the caller goes on to print `$findResults.count`. When only one result is returned, `$findResult` will be a string and not an array. I wonder if the string gets coerced to an array when passed to the Filter method, or whether it also errors.

ericsciple:
@Thilas sorry, this fell off my radar. I have looked over this now and left one comment.

powershell/VstsTaskSdk/FindFunctions.ps1
Thilas: > I see the caller goes on to print `$findResults.count`. When only one result is returned, `$findResult` will be a string and not an array. I wonder if the string gets coerced to an array when passed to the Filter method, or whether it also errors.

Actually, `$findResults.count` is not an issue because even if it is `$null`, it returns `$null`. On the opposite, `$findResults` was not coerced to an array when passed to the Filter method and containing `$null`. That was the bug I had. In such case, it throws a `Value cannot be null. Parameter name: source` probably because of a guard in `Filter` method. See the unit test I added to illustrate that: `Find-Match.SupportsPathNotFound.ps1`.

Otherwise I'm fine with your proposal, the bug I had is still fixed and it seems indeed more powershell spirit than my proposal.

Thilas:
No problem, thanks for looking at it. I just pushed a new commit to comply with your proposal.

Thilas:
@ericsciple thanks!

PR #371, opened 2018-06-19
norc when running tests

ericsciple:


damccorm:
Looks like this was added elsewhere so I'm going to close

PR #373, opened 2018-06-21
Adding an overload of test result publish API to get the test run system from the publisher tasks.

navin22:


bryanmacfarlane:
Looks good to me

PR #382, opened 2018-07-25
Make it more obvious that output is buffered in execSync

keithrob:


node/toolrunner.ts
```diff
@@ -796,6 +796,7 @@ export class ToolRunner extends events.EventEmitter {
         this.args.forEach((arg) => {
             this._debug('   ' + arg);
         });
+        this._debug('IMPORTANT: output from this command is buffered.  There may be a delay between execution and when results are printed.');
```
bryanmacfarlane: debug is for user debugging their definition.  it's not for the author inappropriately using the api.  this has the user seeing output based on the task authors poor decision

keithrob: @bryanmacfarlane ,
I hear what you are saying about "inappropriately using the api"; however, I think that the async version is broken because it does not always print stdout/stderr when the exit code of the spawned command is !=0.  This has caused us a lot of problems because the command will fail and we have to tell the customer to "Please retry with verbose enabled".

I will abandon this PR and switch my task back to using async if we can agree that the async method should always print STDOUT/STDERR when the exit code is !=0. 

All that said, I do think there is some merit in printing the execution time of the spawned command.  That is useful.

bryanmacfarlane: async should always print stdout/stderr if exit code != 0.  note that per OS, stdout is buffered and stderr is not.  We just listen to events and output.  regarding execution time, the logs have command at start and return at end with time stamps.

damccorm:
Looks like this is a no-op so I'm going to close. Feel free to reopen if appropriate.

PR #402, opened 2018-09-27
Include a note on ARM and x86 agent support

juliobbv:
Add a note on ARM and x86 support, as v2.141.0 breaks a the implicit assumption that tasks can only run on x64.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,10 @@ A task which automates Powershell technologies can be written with our Powershel
 
 Documentation: [PowerShell API](powershell/Docs/README.md)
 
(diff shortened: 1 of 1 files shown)


juliobbv:
@TingluoHuang another small PR for you to look at 😄

stephenmichaelf:
Should we consider exposing this as a function that returns an enum to make it easier for task authors?

damccorm:
@stephenmichaelf how about we merge here and follow up on adding the convenience function in #405 (which was created to address that exact issue). Are you comfortable with that?

PR #404, opened 2018-09-30
Rename VSTS to Azure DevOps

pascalberger:
Rename VSTS to Azure DevOps


damccorm:
Looks like this got taken care of when we renamed the repo

PR #374, opened 2018-06-22
Add `getPlatform` function

brcrista:
I've been copy-pasting this in various tasks and feel like it could be a good addition to task-lib.

This improves on the existing `osType` function by using an enum instead of a string (more type safe) and makes it obvious what the possible return values are.

PR #405, opened 2018-10-01
Add CPU architecture to task lib

migheorghe:


damccorm:
@migheorghe are you interested in reviving this? It looks useful, maybe we can do this returning an enum and use the `Agent.OSArchitecture` variable for checking instead though (see [here](https://github.com/Microsoft/azure-pipelines-task-lib/pull/402#issuecomment-425913134) for conversation on this).

I'm thinking something similar to [this](https://github.com/Microsoft/azure-pipelines-task-lib/pull/374), but for architecture.

damccorm:
I'm going to close here for now. If you're interested in reviving this, comment and I'll reopen.

PR #409, opened 2018-10-05
Fix links to MSBuild.ps1 and VSBuild.ps1

jcansdale:
It appears their folders now contain a `V1`. It looks like it's now `MSBuildV1/MSBuild.ps1` and `VSBuildV1/VSBuild.ps1`.


damccorm:
This got taken care of in a different PR so I'm going to close.

PR #413, opened 2018-10-11
Added support for empty string in toolRunner

### LINKED ISSUE #277: ToolRunner should respect empty string args
To un-deprecate a package in npm you're supposed to leave the message string blank. 

When this is done in the npm task it throws an error since it expects a message, but it doesn't receive one. To repro this bug, create a build with two consecutive npm custom command tasks. The first command should be "deprecate <pkg>@<version> "message" ", the second should be "deprecate <pkg>@<version> "" ". 

The problem is in the toolrunner class, specifically _argStringToArray. It should parse two consecutive double-quotes into an empty string, but instead it ignores it.  By adding a check for two consecutive quotes at line 43 this can be avoided.

damccorm:
I still lack the context to be sure that this is a desired change/worth any potential cost in back compat, but this allows users to specify an empty string as an arg for toolrunner. The issue this addresses seems to be a pretty valid use case that I don't see a workaround for.

Fixes #277

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -20,7 +20,7 @@
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
       "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
       "requires": {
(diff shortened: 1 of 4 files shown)


damccorm:
This PR came from a repo that I've since deleted... I added #569 because that's way easier than trying to fix merge conflicts from a non-existent fork 😨

It should be virtually identical, but still probably worth giving a once over @jtpetty before we merge it

PR #414, opened 2018-10-11
Updated task variables to support dots

### LINKED ISSUE #303: Having a dot in a task variable name is not supported
In a task, if you save a variable with a dot in its name, it will be replaced by a '_'.
`tl.setTaskVariable('foo.bar', 'value');`
https://github.com/Microsoft/vsts-task-lib/blob/a4be3729c2416c90fc6876902eb60e5e062d7c63/node/internal.ts#L260

But when trying to read this variable back, the dot is not replaced, so the value is not found.
`tl.getTaskVariable('foo.bar');`
https://github.com/Microsoft/vsts-task-lib/blob/a4be3729c2416c90fc6876902eb60e5e062d7c63/node/task.ts#L420

damccorm:
Fixes #303

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -183,7 +183,7 @@ export interface VariableInfo {
  * @returns   string
  */
 export function getInput(name: string, required?: boolean): string {
-    var inval = im._vault.retrieveSecret('INPUT_' + name.replace(/\./g, '_').replace(/ /g, '_').toUpperCase());
+    var inval = im._vault.retrieveSecret('INPUT_' + im._getVariableKey(name));
     if (inval) {
         inval = inval.trim();
     }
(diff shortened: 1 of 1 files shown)


stephenmichaelf:
There may be other implications, I am not sure we know that any _ came from a .

@ericsciple @TingluoHuang

damccorm:
@stephenmichaelf I don't think what I wrote assumes that any ```_``` came from a ```.``` - it just assumes that if we're trying to get a variable with a ```.``` in its name, then that ```.``` would have been converted to a ```_``` before being stored, which I believe is the done [here](https://github.com/Microsoft/azure-pipelines-task-lib/blob/a4be3729c2416c90fc6876902eb60e5e062d7c63/node/internal.ts#L260)

damccorm:
Migrated to #570 and merged there since this fork got deleted (so I can't resolve conflicts)

PR #416, opened 2018-10-11
Updated vault to encode secrets using safer cipheriv

### LINKED ISSUE #380: Include support for Node 8.7+
Using Node ~8.7+ prints warnings like: `(node:37380) Warning: Use Cipheriv for counter mode of aes-256-ctr`

Because the cipher is hardcoded [here](https://github.com/Microsoft/vsts-task-lib/blob/1195328c3ab17a17a75360015e230193177c4981/node/vault.ts#L8) to an algorithm that later versions of Node warn against.

damccorm:
Currently, if using node version >= 8.7 (and maybe lower) encryption generates a warning. This fixes it and also makes encryption safer (apparently there are flaws with just using createCipher and CTR).

Fixes #380

### DIFF
--- a/node/vault.ts
+++ b/node/vault.ts
@@ -6,6 +6,8 @@ import crypto = require('crypto');
 
 var uuidV4 = require('uuid/v4');
 var algorithm = "aes-256-ctr";
+var encryptEncoding: 'hex' = 'hex';
+var unencryptedEncoding: 'utf8' = 'utf8';
 
 //
 // Store sensitive data in proc.
@@ -43,10 +45,14 @@ export class Vault {
         }
 
         var key = this.getKey();
-        var cipher = crypto.createCipher(algorithm, key);
-        var crypted = cipher.update(data,'utf8','hex')
-        crypted += cipher.final('hex');
-        this._store[name] = crypted;
+        var hashedKey = crypto.createHash('sha256').update(key).digest();
+        var iv = crypto.randomBytes(16);
+
+        var cipher = crypto.createCipheriv(algorithm, hashedKey, iv);
+        var crypted = cipher.update(data, unencryptedEncoding, encryptEncoding)
+        var cryptedFinal = cipher.final(encryptEncoding);
+
+        this._store[name] = iv.toString(encryptEncoding) + crypted + cryptedFinal;
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


node/vault.ts
```diff
@@ -43,10 +45,14 @@ export class Vault {
...
 
         var key = this.getKey();
-        var cipher = crypto.createCipher(algorithm, key);
-        var crypted = cipher.update(data,'utf8','hex')
-        crypted += cipher.final('hex');
-        this._store[name] = crypted;
+        var hashedKey = crypto.createHash('sha256').update(key).digest();
+        var iv = crypto.randomBytes(16);
+
+        var cipher = crypto.createCipheriv(algorithm, hashedKey, iv);
```
ericsciple: will this work in node 5?

node/vault.ts
```diff
@@ -56,11 +62,17 @@ export class Vault {
 
         if (this._store.hasOwnProperty(name)) {
             var key = this.getKey();
+            var hashedKey = crypto.createHash('sha256').update(key).digest();
```
ericsciple: sha512? cj might have a good suggestion

ericsciple: should this move into the getKey function?

node/vault.ts
```diff
@@ -6,6 +6,8 @@ import crypto = require('crypto');
 
 var uuidV4 = require('uuid/v4');
 var algorithm = "aes-256-ctr";
+var encryptEncoding: 'hex' = 'hex';
+var unencryptedEncoding: 'utf8' = 'utf8';
```
ericsciple: lgtm

- let's move the duplicate logic into `getKey`
- check w/ CJ and let's use sha512
- make sure it works on node 5, otherwise fallback to the old logic?

ericsciple: - and discuss with cj the difference between the hash approach and scrypt. I think the hash approach is sufficient based on our usage. I might be worth doing a small perf test between the two. For example, run in a loop N times and measure the total time; repeat several times to make sure the results are consistent.

node/vault.ts
damccorm: Confirmed that it will

damccorm:
Moved to #505 because of branch issues.

PR #417, opened 2018-10-11
export write-logging command for powershell

### LINKED ISSUE #395: Export "Write-LoggingCommand" for Powershell lib
Currently, the Powershell lib doesn't export the `Write-LoggingCommand` which Node.js lib does.

```
// https://github.com/Microsoft/vsts-task-lib/blob/master/node/task.ts#L474
export const command = im._command;
```

It seems the only way to achieve functions like the following is to write the `##vso[]` string manually.
```
// https://github.com/Microsoft/vsts-task-lib/blob/master/node/task.ts#L1717
command('results.publish', properties, '');
command('codecoverage.publish', properties, "");
```

The `Write-LoggingCommand` should be exported so that the same functions can be achieved.

damccorm:
If this is something we want, we should make corresponding changes [here](https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md) once this is in.

Fixes #395

### DIFF
--- a/powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
+++ b/powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
@@ -436,6 +436,41 @@ function Write-UpdateReleaseName {
     Write-LoggingCommand -Area 'release' -Event 'updatereleasename' -Data $Name -AsOutput:$AsOutput
 }
 
+<#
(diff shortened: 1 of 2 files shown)


jtpetty:
LGTM - @vtbassmatt says he's moved docs to another area so we do not need to make the corresponding docs change you referenced above

PR #418, opened 2018-10-11
Don't swallow [command] string in mock-toolrunner

### LINKED ISSUE #212: IExecOptions.silent = true causes MockTestRunner to return 0 executions.
If you run a command with `IExecOptions.silent = true`, the `MockTestRunner` will not think the tool has run. This is because the `MockToolRunner` swallows the `[command]` string with the silent option set. Even if it does not send to whole command to the output, `MockToolRunner` should still send `[command]` so it can be counted.

damccorm:
Fixes #212

### DIFF
--- a/node/mock-toolrunner.ts
+++ b/node/mock-toolrunner.ts
@@ -220,6 +220,9 @@ export class ToolRunner extends events.EventEmitter {
 
             ops.outStream.write('[command]' + cmdString + os.EOL);
         }
+        else {
+            ops.outStream.write('[command]');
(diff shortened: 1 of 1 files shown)


stephenmichaelf:
Not sure this is the right solution to the problem. If people ask for it to be silent shouldn't we be silent? :) Is there another way we can solve the issue?

damccorm:
Yeah, in retrospect I think that I agree. I'll close here and try to find a better solution going forward.

PR #419, opened 2018-10-11
Added setVariableName method

### LINKED ISSUE #148: Add "setVariable(name: string, value: string, isSecret?: boolean)" method to mock-run.ts
In [mock-run.ts class](https://github.com/Microsoft/vsts-task-lib/blob/master/node/mock-run.ts) there is a method to setInput() method. It would be great to have a setVariable method for mocking variables during test. The method should take:
- **name**: name of the variable to add
- **value**: value of the variable
- **isSecret**: optional boolean that sets the variable as a secret if set to true (not secret if false or not supplied)

damccorm:
Resolves #148

### DIFF
--- a/node/mock-run.ts
+++ b/node/mock-run.ts
@@ -16,7 +16,7 @@ export class TaskMockRunner {
     }
 
     public setVariableName(name: string, val: string, isSecret?: boolean) {
-        if(isSecret) {
+        if (isSecret) {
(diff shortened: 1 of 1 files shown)


stephenmichaelf:
Should check why it failed on Ubuntu before merging.

damccorm:
Pretty sure its unrelated, but I'll queue another build to make sure.

damccorm:
Yep, all passing now, I'll go ahead and merge

PR #421, opened 2018-10-12
Fix executable check in _tryGetExecutablePath

### LINKED ISSUE #420: tl.which('toolname') incompatible with dotnet tool install
The which function is calling _tryGetExecutablePath, and that function test for the mode bit 001 to determine if a file is executable

https://github.com/Microsoft/vsts-task-lib/blob/de5f350a063ac2e62da225ea469f224524825277/node/internal.ts#L448

That doesn't work if the executable was created by `dotnet tool install` because the resulting mode bits are 744. Here is an example of installing a tool named `atlas`.

```
ls -la /var/lib/vstsagent/1/_work/r2/a/tools
2018-10-12T18:40:53.4814323Z drwxr-xr-x 3 azureuser azureuser   4096 Oct 12 18:40 .
2018-10-12T18:40:53.4814794Z drwxr-xr-x 3 azureuser azureuser   4096 Oct 12 18:40 ..
2018-10-12T18:40:53.4815183Z -rwxr--r-- 1 azureuser azureuser 106912 Oct 12 18:40 atlas
2018-10-12T18:40:53.4815678Z drwxr-xr-x 4 azureuser azureuser   4096 Oct 12 18:40 .store
```

damccorm:
Right now, this is only checking if the file is executable for anyone (with no special permissions), this ensures that it returns the path if it is executable for the user || for the group || for anyone.

E.g. rwxr--r-- fails right now, it should still return the filepath.

Fixes #420


lodejard:
That was quick! Might want to update this spot also?

https://github.com/Microsoft/vsts-task-lib/blob/de5f350a063ac2e62da225ea469f224524825277/node/internal.ts#L490

node/internal.ts
```diff
@@ -442,10 +442,10 @@ function _tryGetExecutablePath(filePath: string, extensions: string[]): string {
                 }
             }
             else {
-                // on Mac/Linux, test the execute bit
+                // on Mac/Linux, test the execute bits
                 //     R   W  X  R  W X R W X
                 //   256 128 64 32 16 8 4 2 1
-                if ((stats.mode & 1) == 1) {
+                if ((stats.mode & 73) > 0) {
```
ericsciple: This approach seems OK to me, although it's technically inaccurate (slightly - e.g. am i the owner, group member, or everyone?). It might be worth asking Zach if he has any quick ideas.

I'm curious in the case where it's inaccurate, what's the result? How does the error manifest to the user? I want to avoid a future investigation for an error surfacing in a cryptic way.

node/internal.ts
```diff
@@ -487,7 +487,7 @@ function _tryGetExecutablePath(filePath: string, extensions: string[]): string {
                     // on Mac/Linux, test the execute bit
                     //     R   W  X  R  W X R W X
                     //   256 128 64 32 16 8 4 2 1
-                    if ((stats.mode & 1) == 1) {
+                    if ((stats.mode & 73) == 1) {
```
ericsciple: It might be worth a comment out to the right like:

```
if ((stats.mode & 73) == 1) { // 73 = 64 + 8 + 1
```

or more explicit:
```
if ((stats.mode & (64 + 8 + 1)) == 1) {
```

ericsciple: tests?

ericsciple: note, existing tests for `which` are in dirtests.ts

damccorm:
I think this should resolve https://github.com/Microsoft/azure-pipelines-agent/issues/1862 as well

node/internal.ts
```diff
@@ -487,7 +487,7 @@ function _tryGetExecutablePath(filePath: string, extensions: string[]): string {
                     // on Mac/Linux, test the execute bit
                     //     R   W  X  R  W X R W X
                     //   256 128 64 32 16 8 4 2 1
-                    if ((stats.mode & 1) == 1) {
+                    if ((stats.mode & 1) > 0 || ((stats.mode & 8) > 0 && stats.gid === process.getgid()) || ((stats.mode & 64) > 0 && stats.uid === process.getuid())) {
```
ericsciple: neat

node/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "azure-pipelines-task-lib",
-  "version": "2.7.8",
+  "name": "vsts-task-lib",
```
ericsciple: did you mean to change the name?

damccorm: Whoops no, not sure how that happened - good catch.

node/test/dirtests.ts
```diff
@@ -96,6 +96,51 @@ describe('Dir Operation Tests', function () {
 
         done();
     });
+
+    // which tests
+    it('which() finds executable with scoped permissions', function (done) {
```
ericsciple: should this be four tests? one for owner, one for group, one for everyone, one for RW_RW_RW_.

node/internal.ts
```diff
@@ -442,10 +442,10 @@ function _tryGetExecutablePath(filePath: string, extensions: string[]): string {
                 }
             }
             else {
-                // on Mac/Linux, test the execute bit
+                // on Mac/Linux, test the execute bits
                 //     R   W  X  R  W X R W X
                 //   256 128 64 32 16 8 4 2 1
-                if ((stats.mode & 1) == 1) {
+                if ((stats.mode & 1) > 0 || ((stats.mode & 8) > 0 && stats.gid === process.getgid()) || ((stats.mode & 64) > 0 && stats.uid === process.getuid())) {
```
ericsciple: would a private helper method be better since this is used from two locations (single unit test would always excersize the same code)

node/internal.ts
ericsciple: i like that this short-circuits and doesnt need to call `process.getgid()` or `process.getuid()` most of the time

node/test/dirtests.ts
ericsciple: or a foreach that runs the same test, but sets the mode differently on each iteration

ericsciple:
looks good. I left some suggestions.

node/test/dirtests.ts
damccorm: Yeah, this is a good idea. I think I like having separate tests for understandability/descriptive test names rather than a foreach (it can be pulled out into a function though)

node/internal.ts
damccorm: Yep, good call

PR #457, opened 2018-10-15
Fix broken references from rebranding

damccorm:
When we rebranded, a lot of requires and documentation ended up with broken references. Right now, some things are definitely broken because of this, so we should probably publish once this goes through.


bryanmacfarlane:
Looks good.  Most links support 302 but good to clean up.  Thanks! 

Good to merge

bryanmacfarlane:
look into linux failure in checks - doesn't look related.

damccorm:
> Most links support 302 but good to clean up

Agreed on the links - I was trying to use the library as I was creating docs and some things were failing because of bad requires if I didn't install vsts-task-lib though. When I locally changed the references to azure-pipelines-task-lib those errors went away.

PR #459, opened 2018-10-18
Upgrade to TypeScript 3

damccorm:
Pretty straightforward, only changes were due to writeFileSync now apparently requiring an object. Setting a string to the encoding field does the same thing anyways though.


damccorm:
I don't think that this should have a significant impact on how code is compiled to JavaScript, mostly it just lets us do more things (like #233)

damccorm:
@bryanmacfarlane are you ok to move forward with this?

node/task.ts
```diff
@@ -554,7 +554,12 @@ export interface FsOptions {
 }
 
 export function writeFile(file: string, data: string | Buffer, options?: string | FsOptions) {
-    fs.writeFileSync(file, data, options);
+    if(typeof(options) == 'string'){
```
stephenmichaelf: ===?

stephenmichaelf:
For some reason I can't approve. New UX? :)

PR #234, opened 2017-04-04
Have mock-toolrunner emit stdline and errline events.

ILMTitan:
Fix for issue #227.

PR #462, opened 2018-10-22
Mock stdline and errline

### LINKED ISSUE #227: Mock-ToolRunner fails to emit stdline.
ToolRunner will emit to the event 'stdline'. Mock-ToolRunner does not. Tasks that rely on the 'stdline' event are untestable.

damccorm:
Based off of #234 which was closed due to inactivity.

Fixes #227.

### DIFF
--- a/node/test/mocktests.ts
+++ b/node/test/mocktests.ts
@@ -11,6 +11,7 @@ import * as mtr from '../_build/mock-toolrunner';
 import * as ma from '../_build/mock-answer';
 import * as tl from '../_build/task';
 
+import os = require('os');
 import testutil = require('./testutil');
 
 describe('Mock Tests', function () {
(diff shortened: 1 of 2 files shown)


node/test/mocktests.ts
```diff
@@ -183,5 +184,53 @@ describe('Mock Tests', function () {
...
+        });
+        tool.on('stdline', (out) => {
+            if (firstStdline) {
+                assert.equal("atool output here", out);
+                firstStdline = false;
+            }
+            else {
+                assert.equal("abc", out);
+            }
+        });
```
hashtagchris: It would be good to assert this was called exactly twice.

node/test/mocktests.ts
```diff
@@ -183,5 +184,53 @@ describe('Mock Tests', function () {
         
         assert(tool, "tool should not be null");
         assert(rc == 0, "rc is 0");
-    })                
+    })        
+    
+    it('Mock toolRunner returns correct output', async () => {
+        const expectedStdout = "atool output here" + os.EOL + "abc";
```
hashtagchris: optional: You could add a separate test for trailing newline, e.g. `const expectedStdout = os.EOL`

node/test/mocktests.ts
```diff
@@ -183,5 +184,100 @@ describe('Mock Tests', function () {
...
+            numStdLineCalls += 1;
+            assert.equal("", out);
+        });
+        tool.on('errline', (out) => {
+            numStdErrCalls += 1;
+            assert.equal("", out);
+        });
+        await tool.exec(<mtr.IExecOptions>{});
+        
+        assert.equal(numStdLineCalls, 2);
```
hashtagchris: Based on my reading of `processLineBuffer`, I think these would be 1 for the real tool runner.

hashtagchris: Based on my reading of `processLineBuffer`, I think these would be 1 for the real tool runner.

damccorm: Oh whoops, you're right. Right now I'm emitting all elements of ```res.stderr.split(os.EOL)``` but really I should be ignoring the last one, that's what is causing this issue. Good catch.

damccorm: Should be fixed now

node/test/mocktests.ts
```diff
@@ -183,5 +184,100 @@ describe('Mock Tests', function () {
         
         assert(tool, "tool should not be null");
         assert(rc == 0, "rc is 0");
-    })                
+    })
+    
+    it('Mock toolRunner returns correct output', async () => {
+        const expectedStdout = "atool output here" + os.EOL + "abc" + os.EOL + "junk";
```
hashtagchris: Wouldn't/shouldn't the real task runner emit 'stdline' for "junk"?

hashtagchris: https://github.com/Microsoft/azure-pipelines-task-lib/blob/2f0dde7e1786bcca01f0e8a0d6bdaa760b89ea8f/node/toolrunner.ts#L667

damccorm: Yes, and it does [on close](https://github.com/Microsoft/azure-pipelines-task-lib/blob/2f0dde7e1786bcca01f0e8a0d6bdaa760b89ea8f/node/toolrunner.ts#L666). Sorry, looks like my brain didn't quite get going this morning :)

damccorm: Should actually be fixed now

hashtagchris:
Should the version number in node/package.json be bumped?

PR #463, opened 2018-10-22
Mock getInputs

### LINKED ISSUE #372: mock-task can't load task inputs and variables when AGENT_TEMPDIRECTORY is set
Apparently from this change: https://github.com/Microsoft/vsts-task-lib/blame/master/node/internal.ts#L692

When tasks have tests that set a value for AGENT_TEMPDIRECTORY, the `getInput` function on mock-task (which is the same as the production version and calls into that _loadData function) is unable to load any of the task inputs.

**Repro**
install version >= 2.1 of task-lib to one of the following tasks and run the tests:
* DotNetCoreCLIV2
* InstallAppleCertificateV2
* InstallSSHKeyV0

damccorm:
Currently, the mock of getInputs fails if the base directory is changed since setInputs doesn't adjust for this and getInputs is not mocked. This PR mocks getInputs.

Fixes #372

### DIFF
--- a/node/mock-task.ts
+++ b/node/mock-task.ts
@@ -4,6 +4,7 @@ import path = require('path');
 import fs = require('fs');
 import os = require('os');
 import util = require('util');
+import im = require('./internal');
 import task = require('./task');
 import tcm = require('./taskcommand');
 import trm = require('./mock-toolrunner');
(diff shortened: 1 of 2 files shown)


brcrista:
I have a change planned that will let `setAnswers` mock arbitrary task lib functions.  I ran it past @bryanmacfarlane , @ericsciple , and @stephenmichaelf a few months ago.  Haven't scheduled time for it yet though

damccorm:
Closing because I don't think this will actually work since `_loadData` [deletes env variables](https://github.com/Microsoft/azure-pipelines-task-lib/blob/1a7e4ddca5277f87301b4880fed1971022c16b7e/node/internal.ts#L723).

I like the idea of letting `setAnswers` mock task-lib functions, might still be good to have a better default mock here though.

PR #465, opened 2018-10-25
Special case Microsoft.TeamFoundation.Build.WebApi.*

### LINKED ISSUE #263: Get-VstsVssHttpClient assembly load fails for Microsoft.TeamFoundation.Build.WebApi clients
Hi,

I am trying to instantiate Microsoft.TeamFoundation.Build.WebApi.XamlBuildHttpClient (from assembly [Microsoft.TeamFoundation.Build2.WebApi](https://www.nuget.org/packages/Microsoft.TeamFoundationServer.Client/)) via Get-VstsVssHttpClient function, and I get "Cannot convert the ... value of type System.String to type System.Type" error.

I found similar [Issue 97](https://github.com/Microsoft/vsts-task-lib/issues/197), but I'm already placing all required assemblies in the root of my task.
I've managed to resolve the issue temporarily by renaming Microsoft.TeamFoundation.Buil**d2.W**ebApi.dll into Microsoft.TeamFoundation.Buil**d.W**ebApi.dll
Maybe you can improve your assembly finding [mechanisms](https://github.com/Microsoft/vsts-task-lib/blob/master/powershell/VstsTaskSdk/ServerOMFunctions.ps1#L427).

damccorm:
Resolves #263

### DIFF
--- a/powershell/VstsTaskSdk/ServerOMFunctions.ps1
+++ b/powershell/VstsTaskSdk/ServerOMFunctions.ps1
@@ -601,7 +601,7 @@ function Get-OMType {
             if ($namespace -like 'Microsoft.TeamFoundation.Build.WebApi.*' -or
                 $namespace -eq 'Microsoft.TeamFoundation.Build.WebApi') {
 
-                $build2Namespace = 'Microsoft.TeamFoundation.Build2' + $namespace.Substring(30)
+                $build2Namespace = 'Microsoft.TeamFoundation.Build2' + $namespace.Substring( 'Microsoft.TeamFoundation.Build'.Length)
                 $dllPaths += [System.IO.Path]::Combine($OMDirectory, "$build2Namespace.dll")
             }
         }
--- a/powershell/VstsTaskSdk/ServerOMFunctions.ps1
+++ b/powershell/VstsTaskSdk/ServerOMFunctions.ps1
@@ -597,12 +597,12 @@ function Get-OMType {
                     $dllPaths += [System.IO.Path]::Combine($OMDirectory, "$namespace.WebApi.dll")
                     $dllPaths += [System.IO.Path]::Combine($OMDirectory, "$namespace.dll")
                 }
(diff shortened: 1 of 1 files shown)


powershell/VstsTaskSdk/ServerOMFunctions.ps1
```diff
@@ -598,6 +598,12 @@ function Get-OMType {
                     $dllPaths += [System.IO.Path]::Combine($OMDirectory, "$namespace.dll")
                 }
             }
+            if ($namespace -like 'Microsoft.TeamFoundation.Build.WebApi.*' -or
+                $namespace -eq 'Microsoft.TeamFoundation.Build.WebApi') {
+
+                $build2Namespace = 'Microsoft.TeamFoundation.Build2' + $namespace.Substring(30)
```
ericsciple: nit: consider `'Microsoft.TeamFoundation.Build'.Length` over `$namespace.Substring(30)`

powershell/VstsTaskSdk/ServerOMFunctions.ps1
```diff
@@ -598,6 +598,12 @@ function Get-OMType {
                     $dllPaths += [System.IO.Path]::Combine($OMDirectory, "$namespace.dll")
                 }
             }
+            if ($namespace -like 'Microsoft.TeamFoundation.Build.WebApi.*' -or
```
ericsciple: I think this should be moved into the extended client block

ericsciple: actually i'm not sure which block it should be moved into, but I think it should be moved into one of them. which one was the use case?

damccorm: I think it was from the WebApi, though its a little bit unclear since the project has since moved to a TypeScript implementation. Not 100% sure I understand the rationale for moving it into the blocks anyways though - couldn't this cause an issue with either?

damccorm:
Closing due to lack of traction and lack of customer need. If you need this please comment and we'll reconsider

cc/ @jtpetty

PR #466, opened 2018-10-25
Updated dependencies from component governance

damccorm:
Most issues were coming from an old version of mocha in powershell


stephenmichaelf:
Could you add some context, what's this for and how does it relate to CG? I just see dev dependencies updated. Why did it start failing?

damccorm:
It looks like PowerShell can't find minimatch, which is weird, I can't repro on my machine. Could be a node versioning issue, I think I'm using node 8 instead of 6, I'll dig into it further.

For context, some of the dependencies have security issues, by updating we fix them. For most of the dependencies (e.g. mocha), the security issue is caused by something further down the dependency tree, so its hard to see exactly what the issue is, you can look at it by running npm audit though.

damccorm:
Looks like adm-zip was the culprit that was causing issues, I'm just going to leave it since its just a dev dependency in make-options and not really an issue. We can at least get rid of the other warnings this way though.

Issue #476, opened 2018-11-26
No command wrappers in node lib


mmajcica:
In powershell library we have plenty of the wrapper that do make code cleaner and easier to use, like Write-VstsUploadFile
https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/powershell/Docs/Commands.md#write-vstsuploadfile

In Node lib this is missing and it would make the code I'm writing way cleaner.
I was wondering, would a PR with some of the commands be welcome?

damccorm:
Sure, that sounds like a good improvement!

damccorm:
Fixed by #477

PR #477, opened 2018-11-27
Added missing logging commands to Node lib

mmajcica:
As proposed in #476 this are the missing commands that the NodeJs lib does not expose.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -1904,7 +1904,7 @@ export function setProgress(percent: number, currentOperation: string) {
 export function logDetail(id: string, parentId?: string, recordType?: string,
     recordName?: string, order?: number, startTime?: string, finishTime?: string,
     progress?: number, state?: TaskState, result?: TaskResult, message?: string) {
-    const properties = { "id": id };
+    var properties = { "id": id };
 
     if (parentId) {
         properties["parentid"] = parentId;
@@ -1958,7 +1958,7 @@ export function logDetail(id: string, parentId?: string, recordType?: string,
  */
 export function logIssue(type: IssueType, sourcePath: string, lineNumber: number,
     columnNumber: number, errorCode: string, message: string) {
-    const properties = { "type": type };
+    var properties = { "type": type };
 
     if (errorCode) {
         properties["code"] = errorCode;
@@ -1994,8 +1994,14 @@ export function logIssue(type: IssueType, sourcePath: string, lineNumber: number
  * @param path              Path to the file that should be uploaded.
  * @returns                 void
  */
-export function uploadArtifact(containerFolder: string, name: string, path: string) {
-    command("artifact.upload", { "containerfolder": containerFolder, "artifactname": name }, path);
+export function uploadArtifact(containerFolder: string, path: string, name?: string) {
+    var properties = { "containerfolder": containerFolder };
+    
+    if (name) {
+        properties["artifactname"] = name;
+    }
+
+    command("artifact.upload", properties, path);
 }
 
 /**
--- a/node/task.ts
+++ b/node/task.ts
@@ -1901,9 +1901,9 @@ export function setProgress(percent: number, currentOperation: string) {
  * @param message       current operation
  * @returns             void
  */
-export function logDetail(id: string, parentId?: string, recordType?: string,
+export function logDetail(id: string,  message: string, parentId?: string, recordType?: string,
     recordName?: string, order?: number, startTime?: string, finishTime?: string,
-    progress?: number, state?: TaskState, result?: TaskResult, message?: string) {
+    progress?: number, state?: TaskState, result?: TaskResult) {
     var properties = { "id": id };
 
     if (parentId) {
@@ -1996,7 +1996,7 @@ export function logIssue(type: IssueType, sourcePath: string, lineNumber: number
  */
 export function uploadArtifact(containerFolder: string, path: string, name?: string) {
     var properties = { "containerfolder": containerFolder };
-    
+
     if (name) {
         properties["artifactname"] = name;
     }
--- a/node/task.ts
+++ b/node/task.ts
@@ -1901,46 +1901,21 @@ export function setProgress(percent: number, currentOperation: string) {
  * @param message       current operation
  * @returns             void
  */
-export function logDetail(id: string,  message: string, parentId?: string, recordType?: string,
+export function logDetail(id: string, message: string, parentId?: string, recordType?: string,
     recordName?: string, order?: number, startTime?: string, finishTime?: string,
     progress?: number, state?: TaskState, result?: TaskResult) {
-    var properties = { "id": id };
-
-    if (parentId) {
-        properties["parentid"] = parentId;
-    }
-
-    if (recordType) {
-        properties["type"] = recordType;
-    }
-
-    if (recordName) {
-        properties["name"] = recordName;
-    }
-
-    if (order) {
-        properties["order"] = order.toString();
-    }
-
-    if (startTime) {
-        properties["starttime"] = startTime;
-    }
-
-    if (finishTime) {
-        properties["finishtime"] = finishTime;
-    }
-
-    if (progress) {
-        properties["progress"] = progress.toString();
-    }
-
-    if (state) {
-        properties["state"] = TaskState[state];
-    }
-
-    if (result) {
-        properties["result"] = TaskResult[result];
-    }
+    const properties = {
+        "id": id,
+        "parentid": parentId,
+        "type": recordType,
+        "name": recordName,
+        "order": order ? order.toString() : undefined,
+        "starttime": startTime,
+        "finishtime": finishTime,
+        "progress": progress ? progress.toString() : undefined,
+        "state": state ? TaskState[state] : undefined,
+        "result": result ? TaskResult[result] : undefined
+    };
 
     command("task.logdetail", properties, message);
 }
@@ -1956,25 +1931,15 @@ export function logDetail(id: string,  message: string, parentId?: string, recor
  * @param message       Error or warning message.
  * @returns             void
  */
-export function logIssue(type: IssueType, sourcePath: string, lineNumber: number,
-    columnNumber: number, errorCode: string, message: string) {
-    var properties = { "type": type };
-
-    if (errorCode) {
-        properties["code"] = errorCode;
-    }
-
-    if (sourcePath) {
-        properties["sourcepath"] = sourcePath;
-    }
-
-    if (lineNumber) {
-        properties["linenumber"] = lineNumber;
-    }
-
-    if (columnNumber) {
-        properties["columnnumber"] = columnNumber;
-    }
+export function logIssue(type: IssueType, message: string, sourcePath?: string, lineNumber?: number,
+    columnNumber?: number, errorCode?: string, ) {
+    const properties = {
+        "type": type,
+        "code": errorCode,
+        "sourcepath": sourcePath,
+        "linenumber": lineNumber ? lineNumber.toString() : undefined,
+        "columnnumber": columnNumber ? columnNumber.toString() : undefined,
+    };
 
     command("task.logissue", properties, message);
 }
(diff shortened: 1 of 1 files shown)


node/task.ts
```diff
@@ -1775,6 +1803,218 @@ export class CodeCoverageEnabler {
...
+
+/**
+ * Update release name for current release.
+ * 
+ * @param value     Value to be assigned as the release name.
+ * @returns         void
+ */
+export function updateReleaseName(name: string) {
+    assertAgent("2.132");
+    command("build.updatereleasename", null, name);
```
damccorm: Shouldn't this be ```"release.updatereleasename"```? https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md#release-logging-commands

node/task.ts
```diff
@@ -1775,6 +1803,218 @@ export class CodeCoverageEnabler {
...
+ * 
+ * The file shall be available for download along with task logs.
+ * 
+ * @param name              Artifact name.
+ * @param path              Path to the file that should be associated.
+ * @param artifactType      ArtifactType enum of Container, FilePath, VersionControl, GitRef or TfvcLabel.
+ * @returns                 void
+ */
+export function associateArtifact(name: string, path: string, artifactType: ArtifactType) {
+    command("artifact.associate", { "type": artifactType, "artifactname": name }, path);
```
damccorm: Nit: It would probably be good to group these functions by type. So everything with the form ```artifact.x``` goes together, everything with the form ```task.x``` goes together, etc...

node/task.ts
```diff
@@ -1775,6 +1803,218 @@ export class CodeCoverageEnabler {
...
+/**
+ * Upload and attach summary markdown to current timeline record.
+ * This summary shall be added to the build/release summary and 
+ * not available for download with logs.
+ * 
+ * @param path      Local directory path.
+ * @returns         void
+ */
+export function uploadSummary(path: string) {
+    command("task.uploadsummary", null, path);
```
damccorm: ```assertAgent("0.5.6");``` https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md

node/task.ts
```diff
@@ -1775,6 +1803,218 @@ export class CodeCoverageEnabler {
...
+ * @param order         Order of timeline record.
+ * @param startTime     Start time.
+ * @param finishTime    End time.
+ * @param progress      Percentage of completion.
+ * @param state         TaskState enum of Unknown, Initialized, InProgress or Completed.
+ * @param result        TaskResult enum of Succeeded, SucceededWithIssues, Failed, Cancelled or Skipped.
+ * @param message       current operation
+ * @returns             void
+ */
+export function logDetail(id: string, parentId: string, recordType: string,
```
damccorm: Most of these properties should be optional - see here https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md

node/task.ts
```diff
@@ -1775,6 +1803,218 @@ export class CodeCoverageEnabler {
...
+ * @param type          IssueType enum of Error or Warning.
+ * @param sourcePath    Source file location.
+ * @param lineNumber    Line number.
+ * @param columnNumber  Column number.
+ * @param code          Error or warning code.
+ * @param message       Error or warning message.
+ * @returns             void
+ */
+export function logIssue(type: IssueType, sourcePath: string, lineNumber: number,
+    columnNumber: number, errorCode: string, message: string) {
```
damccorm: Most of these should be optional - https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md

node/task.ts
```diff
@@ -1775,6 +1803,218 @@ export class CodeCoverageEnabler {
...
+ * Upload user interested file as additional log information 
+ * to the current timeline record.
+ * 
+ * The file shall be available for download along with task logs.
+ * 
+ * @param path      Path to the file that should be uploaded.
+ * @returns         void
+ */
+export function uploadFile(path: string) {
+    command("task.uploadfile", null, path);
```
damccorm: Should have ```assertAgent("1.101")```

node/task.ts
```diff
@@ -1775,6 +1803,218 @@ export class CodeCoverageEnabler {
...
+ * to the current timeline record.
+ * 
+ * The file shall be available for download along with task logs.
+ * 
+ * @param containerFolder   Folder that the file will upload to, folder will be created if needed.
+ * @param name              Artifact name.
+ * @param path              Path to the file that should be uploaded.
+ * @returns                 void
+ */
+export function uploadArtifact(containerFolder: string, name: string, path: string) {
```
damccorm: name should be optional - https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md#artifact-logging-commands

node/task.ts
```diff
@@ -1775,6 +1803,218 @@ export class CodeCoverageEnabler {
...
+    command("build.uploadlog", null, path);
+}
+
+/**
+ * Update build number for current build.
+ * 
+ * @param value     Value to be assigned as the build number.
+ * @returns         void
+ */
+export function updateBuildNumber(value: string) {
```
damccorm: ```assertAgent("1.88")```

node/task.ts
```diff
@@ -1775,6 +1803,218 @@ export class CodeCoverageEnabler {
...
+}
+
+/**
+ * Add a tag for current build.
+ * 
+ * @param value     Tag value.
+ * @returns         void
+ */
+export function addBuildTag(value: string) {
+    command("build.addbuildtag", null, value);
```
damccorm: ```assertAgent("1.95");```

node/task.ts
mmajcica: Indeed, a copy/paste inheritance is paying off. :)
Fixed.

node/task.ts
mmajcica: Group in the code in kind of sections or in classes that are then exported in the task module?

damccorm: Just in sections. Mostly just a style thing to make functions easier to find

node/task.ts
mmajcica: The problem is that you can't assert the agent version lower than 2.104.1
https://github.com/Microsoft/azure-pipelines-task-lib/blob/9fdd82cd724c61aa1615e732a20d3a5ffce67397/node/task.ts#L90

node/task.ts
mmajcica: The problem is that you can't assert the agent version lower than 2.104.1

node/task.ts
damccorm: Oh whoops, my mistake. Good call

node/task.ts
mmajcica: Can you check if the new changes are as expected?

damccorm: Likewise, this can just be:
```
export function logIssue(type: IssueType, message: string, sourcePath?: string, lineNumber?: number,
    columnNumber?: number, errorCode?: string) {
    const properties = {
        "type": type,
        "code": errorCode,
        "sourcepath": sourcePath,
        "linenumber": lineNumber,
        "columnnumber": columnNumber
    };

    command("task.logissue", properties, message);
}
```

node/task.ts
```diff
@@ -1775,6 +1803,271 @@ export class CodeCoverageEnabler {
...
+ * @param progress      Percentage of completion.
+ * @param state         TaskState enum of Unknown, Initialized, InProgress or Completed.
+ * @param result        TaskResult enum of Succeeded, SucceededWithIssues, Failed, Cancelled or Skipped.
+ * @param message       current operation
+ * @returns             void
+ */
+export function logDetail(id: string, parentId?: string, recordType?: string,
+    recordName?: string, order?: number, startTime?: string, finishTime?: string,
+    progress?: number, state?: TaskState, result?: TaskResult, message?: string) {
+    var properties = { "id": id };
```
damccorm: I probably should have been clearer here - we really only need to make this optional in the function declaration, the [check here](https://github.com/Microsoft/azure-pipelines-task-lib/blob/9fdd82cd724c61aa1615e732a20d3a5ffce67397/node/taskcommand.ts#L34) should take care of the rest. So this can still just be
```
export function logDetail(id: string, message: string, parentId?: string, recordType?: string,
    recordName?: string, order?: number, startTime?: string, finishTime?: string,
    progress?: number, state?: TaskState, result?: TaskResult) {
    const properties = {
        "id": id,
        "parentid": parentId,
        "type": recordType,
        "name": recordName,
        "order": order.toString(),
        "starttime": startTime,
        "finishtime": finishTime,
        "progress": progress.toString(),
        "state": TaskState[state],
        "result": TaskResult[result]
    };

    command("task.logdetail", properties, message);
}
```

Sorry, I know I'm being a little picky on style

node/task.ts
```diff
@@ -1775,6 +1803,271 @@ export class CodeCoverageEnabler {
...
+ * 
+ * The file shall be available for download along with task logs.
+ * 
+ * @param containerFolder   Folder that the file will upload to, folder will be created if needed.
+ * @param name              Artifact name.
+ * @param path              Path to the file that should be uploaded.
+ * @returns                 void
+ */
+export function uploadArtifact(containerFolder: string, path: string, name?: string) {
+    var properties = { "containerfolder": containerFolder };
```
damccorm: And this can just be:
```
var properties = { "containerfolder": containerFolder,  "artifactname": name };
```

node/task.ts
mmajcica: You are right to be picky about it.
I initially tried what you suggest, however, non passed parameters will be set to undefined, thus I was scared on how will a undefined string (in some cases) be processed later by the engine.

Check this:
image

mmajcica: Also, in case of the log detail, is the message also optional?

damccorm: I guess you could have a logDetail without a message, I think it probably makes sense to leave it as required though and the user can pass an explicit null or empty string if they want to omit it. My reasoning here is that unlike ```properties```, [there is typing](https://github.com/Microsoft/azure-pipelines-task-lib/blob/9fdd82cd724c61aa1615e732a20d3a5ffce67397/node/internal.ts#L277) on ```message``` in the ```command``` function. I think this might cause issues if strict type checking is turned on (though I'm not 100% sure).

mmajcica: Even if undefined may be handled correctly, in case of strings, a check needs to be preformed first, otherwise a type request ([key: string]: string) will contain a string entry "undefined" which may be assumed later to be the value for a given key.

damccorm: Ah, I see the issue, good catch! Maybe to fix this we can do something like:
```
export function logDetail(id: string, message: string, parentId?: string, recordType?: string,
    recordName?: string, order?: number, startTime?: string, finishTime?: string,
    progress?: number, state?: TaskState, result?: TaskResult) {
    const properties = {
        "id": id,
        "parentid": parentId,
        "type": recordType,
        "name": recordName,
        "order": order ? order.toString() : undefined,
        "starttime": startTime,
        "finishtime": finishTime,
        "progress": progress ? progress.toString() : undefined,
        "state": state ? TaskState[state] : undefined,
        "result": result ? TaskResult[result] : undefined
    };

    command("task.logdetail", properties, message);
}
```

That way any undefined values map correctly to an undefined value in the dictionary and are ignored by ```toString```? I'm mostly just looking for a way to make what is happening a little simpler/less verbose.

mmajcica: I like your solution better, I'll fix it right away.

node/task.ts
mmajcica: done

mmajcica:
@damccorm thank you for accepting it.
This will make my task code cleaner and easier to maintain in the future.
When can we expect this to end up in npm?

damccorm:
@bryanmacfarlane can we release? Changes should be low risk, they include this one, [a small cleanup to fix enum issues here](https://github.com/Microsoft/azure-pipelines-task-lib/commit/e0a9634ca6aa7c0095dc5fbf3608ea2539d430fa), documentation update, and your unc arg change.

bryanmacfarlane:
2.7.7 just published

PR #479, opened 2018-12-03
Updating schema

jessehouwing:


tasks.schema.json
```diff
@@ -32,6 +32,10 @@
     "preview": {
       "type": "boolean"
     },
+    "deprecated": {
+      "type": "boolean",
+      "description": "We consider that a certain task is deprecated only when the latest version of that task is marked as deprecated. When the user searches for deprecated tasks, we push these tasks to the end and group them under a collapsible section that's collapsed by default."
```
damccorm: Nit: This feels pretty wordy/runs pretty far off the screen. Maybe "Task is deprecated only when the latest version is marked as deprecated. Deprecated tasks appear at the end of searches under a section that is collapsed by default."

jessehouwing: Updated!

damccorm: Looks great, thanks for contribution!

PR #483, opened 2018-12-14
Fix infinite loop

### LINKED ISSUE #482: _loadLocStrings runs into an infinite loop
I am using webpack@4.37.1 to package my task dependencies and I am running into an infinite loop when I am trying to run the task locally. 

Most likely it's a faulty configuration on my side but even then the library should not run into infinite loops. I have attached relevant images. 

image

image

damccorm:
Right now, _loc and _loadLocStrings can get into an infinite loop (as described [here](https://github.com/Microsoft/azure-pipelines-task-lib/issues/482#issuecomment-447336556)), this keeps that from happening.

Fixes #482

### DIFF
--- a/node/internal.ts
+++ b/node/internal.ts
@@ -132,7 +132,7 @@ function _loadLocStrings(resourceFile: string, culture: string): { [key: string]
         }
     }
     else {
-        _warning(_loc('LIB_ResourceFileNotExist', resourceFile));
+        _warning('LIB_ResourceFile does not exist');
     }
 
     return locStrings;
@@ -182,12 +182,12 @@ export function _loc(key: string, ...param: any[]): string {
     if (!_libResourceFileLoaded) {
         // merge loc strings from azure-pipelines-task-lib.
         var libResourceFile = path.join(__dirname, 'lib.json');
-        _libResourceFileLoaded = true;
         var libLocStrs = _loadLocStrings(libResourceFile, _resourceCulture);
         for (var libKey in libLocStrs) {
             //cache azure-pipelines-task-lib loc string
             _locStringCache[libKey] = libLocStrs[libKey];
         }
+        _libResourceFileLoaded = true;
     }
 
     var locString;;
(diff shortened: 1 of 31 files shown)


node/internal.ts
```diff
@@ -182,13 +182,12 @@ export function _loc(key: string, ...param: any[]): string {
     if (!_libResourceFileLoaded) {
         // merge loc strings from azure-pipelines-task-lib.
         var libResourceFile = path.join(__dirname, 'lib.json');
+        _libResourceFileLoaded = true;
```
stephenmichaelf: Setting to true before doing anything?

stephenmichaelf: Can we add tests for this?

damccorm: > Setting to true before doing anything?

Yeah, right now if [_loc](https://github.com/Microsoft/azure-pipelines-task-lib/blob/b9602e028255065aed92c64fa5c07f235542b1bc/node/internal.ts#L181) is called for the first time, it calls [_loadLocStrings](https://github.com/Microsoft/azure-pipelines-task-lib/blob/b9602e028255065aed92c64fa5c07f235542b1bc/node/internal.ts#L185) before setting it to true. This normally makes sense, but if [_exist(resourceFile)](https://github.com/Microsoft/azure-pipelines-task-lib/blob/b9602e028255065aed92c64fa5c07f235542b1bc/node/internal.ts#L102) returns false, we [call _loc](https://github.com/Microsoft/azure-pipelines-task-lib/blob/b9602e028255065aed92c64fa5c07f235542b1bc/node/internal.ts#L135) again which kicks off the loop. By setting it to true we avoid this loop, and this is the only place its checked so changing it early shouldn't have any side effects. Basically we're saying that since we've started trying to load libResourceFile we're not going to try to do it again.

As I think about it though, it probably makes more sense to just not call _loc [here](https://github.com/Microsoft/azure-pipelines-task-lib/blob/b9602e028255065aed92c64fa5c07f235542b1bc/node/internal.ts#L135) since its guaranteed not to be able to find the correct localization anyways. I'll update accordingly.

damccorm:
@stephenmichaelf I ran into this issue again while experimenting with webpack - can we merge?

stephenmichaelf:
To confirm, this change just changes 1 line, right?

damccorm:
Yeah. Basically we're trying to localize an error that says that we can't do localization, and that's causing the infinite loop.

PR #487, opened 2019-01-04
Update mock-task with new functions

jotaylo:


damccorm:
I'm not really sure it makes sense to automatically mock all of these functions like this, because the task implementations of these functions assume that they are being run in the context of an agent.

Take assertAgent as an example. assertAgent relies on Agent.Version being set, which the agent does automatically, and shouldn't be the user's responsibility. IMO it would be weird behavior, therefore, to automatically mock assertAgent such that if the test writer doesn't set Agent.Version the task fails with the error message ```Agent version ${minimum} or higher is required```. I think the current behavior of failing and notifying the writer that its because assertAgent isn't mocked actually makes more sense.

To use another example, setSecret relies on [_command](https://github.com/Microsoft/azure-pipelines-task-lib/blob/045cdebde0cc14897ddcacd8de64a51c5898a975/node/internal.ts#L277) which works by writing output that the agent then reads. If this isn't run in the context of an agent, it won't really work. Again, I think the current behavior makes more sense.

In general, I'd say this is pretty consistently true for all of the functions you're trying to mock here (though I haven't examined all of them closely). I'd be open to a contribution that mocks these in a way that captures their functionality better, but I don't think we can just call the task functions to mock them.

damccorm:
I'm going to close since I don't think we should accept this one, @jotaylo feel free to comment here/reopen if you have additional thoughts.

PR #496, opened 2019-02-13
Upgrade to shelljs 0.8

damccorm:
I need to do a little more research to make sure this doesn't have undesirable side effects, but this should allow us to solve https://github.com/Microsoft/azure-pipelines-tasks/issues/5472. If we don't do this we need to figure out a different way to solve the symlink issue.

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -20,7 +20,7 @@
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
       "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
       "requires": {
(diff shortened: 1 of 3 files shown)


damccorm:
After some more digging I feel pretty comfortable making this change, I think the only significant change to the api that we use is that move now has a default of `-f` instead of `-n`, but this PR already handles that for back compat. We have pretty good test coverage which helps.

bryanmacfarlane:
You should chat with Eric and I.  We looked at it before and we were convinced it would break compat / existing tasks.

damccorm:
Interesting, lets touch base later today.

damccorm:
Touched base with Eric, closing. We should consider removing the shelljs dependency when we release the next major version.

damccorm:
Or if we don't want to remove it, we can at least bump the dependency to `0.8.0` with the new major version. That risks breaking tasks that upgrade to consume the new task-lib, but it also seems like it would only be breaking because the functions would be doing what they're supposed to instead of their current behavior.

PR #497, opened 2019-02-14
Fix an small typo in the documentation

judavi:


damccorm:
Good catch, thanks for the contribution!

PR #502, opened 2019-02-27
Allow testing with handlers other than node 10

damccorm:
I still don't love my approach here, but this should at least enable us to handle most of this automatically.

I would love any thoughts on a clean way to (a) find existing/relevant `node.exe` instances that's better than using the `env` and (b) find the `task.json` such that we don't need the user to pass it in (alternately find another way to get the desired node version). I didn't want to search for either of those things since it would need to happen for every test run and it could be really expensive since we don't have any guarantees about where those things would be (this is especially true of the node versions).

### DIFF
--- a/node/mock-test.ts
+++ b/node/mock-test.ts
@@ -1,8 +1,8 @@
 import cp = require('child_process');
 import fs = require('fs');
 import ncp = require('child_process');
-import path = require('path');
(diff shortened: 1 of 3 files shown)


damccorm:
Closing for now, will reopen with new solution discussed offline

PR #503, opened 2019-03-01
Detect correct node handler

damccorm:
This has been fixed up and should be ready for review.

### DIFF
--- a/node/docs/nodeVersioning.md
+++ b/node/docs/nodeVersioning.md
@@ -16,5 +16,5 @@ If this version of node is not found on the path, the library downloads the appr
 
 ### Behavior overrides
 
-To specify a specific version of node to use, set the `useNodeVersion` environment variable to the integer major version (e.g. `process.env['useNodeVersion'] = 5`).
-To specify the location of a `task.json` file, set the `taskJsonPath` environment variable to the path of the file (e.g. `process.env['taskJsonPath'] = path.join(__dirname, 'taskJsonFolder', 'task.json')`).
+To specify a specific version of node to use, set the `nodeVersion` optional parameter in the `run` function of the `MockTestRunner` to the integer major version (e.g. `mtr.run(5)`).
+To specify the location of a `task.json` file, set the `taskJsonPath` optional parameter in the `MockTestRunner` constructor to the path of the file (e.g. `let mtr = new mt.MockTaskRunner('<pathToTest>', '<pathToTask.json>'`).
--- a/node/docs/nodeVersioning.md
+++ b/node/docs/nodeVersioning.md
@@ -5,10 +5,10 @@
 The agent currently has 2 different node handlers that it can use to execute node tasks: Node 6 and Node 10.
 The handler used depends on the `execution` property specified in the tasks `task.json`.
 If the `execution` property is specified to be `Node`, the task will run on the Node 6 handler, if it is specified to be `Node10` it will run on the Node 10 handler.
-Some legacy agents will use a Node 5 handler if the `execution` is specified to be `Node`, but this is a very small percentage of existing agents.
 
 ## Mock-test Node Handler
 
+[Unit testing](https://docs.microsoft.com/en-us/azure/devops/extend/develop/add-build-task?view=azure-devops#step-2-unit-testing-your-task-scripts) of tasks can be done using the task-lib's built in mock-task functionality.
 To ensure tests are run in the same environment as the agent, this library looks for a `task.json` file in the same directory as the supplied task entry point.
 If no `task.json` is found it searches all ancestor directories as well.
 If the `task.json` is still not found, the library defaults to node 10, otherwise it uses the appropriate handler based on the `execution` property.
--- a/node/docs/nodeVersioning.md
+++ b/node/docs/nodeVersioning.md
@@ -1,0 +1,20 @@
+# Node versioning
+
+## Agent Node Handler
+
+The agent currently has 2 different node handlers that it can use to execute node tasks: Node 6 and Node 10.
+The handler used depends on the `execution` property specified in the tasks `task.json`.
+If the `execution` property is specified to be `Node`, the task will run on the Node 6 handler, if it is specified to be `Node10` it will run on the Node 10 handler.
+Some legacy agents will use a Node 5 handler if the `execution` is specified to be `Node`, but this is a very small percentage of existing agents.
+
+## Mock-test Node Handler
+
+To ensure tests are run in the same environment as the agent, this library looks for a `task.json` file in the same directory as the supplied task entry point.
+If no `task.json` is found it searches all ancestor directories as well.
+If the `task.json` is still not found, the library defaults to node 10, otherwise it uses the appropriate handler based on the `execution` property.
+If this version of node is not found on the path, the library downloads the appropriate version.
+
+### Behavior overrides
+
+To specify a specific version of node to use, set the `useNodeVersion` environment variable to the integer major version (e.g. `process.env['useNodeVersion'] = 5`).
+To specify the location of a `task.json` file, set the `taskJsonPath` environment variable to the path of the file (e.g. `process.env['taskJsonPath'] = path.join(__dirname, 'taskJsonFolder', 'task.json')`).
--- a/node/mock-test.ts
+++ b/node/mock-test.ts
@@ -10,7 +10,7 @@ import syncRequest = require('sync-request');
 
 const COMMAND_TAG = '[command]';
 const COMMAND_LENGTH = COMMAND_TAG.length;
-const testDirectory = path.join(__dirname, '_test');
+const downloadDirectory = path.join(process.env.HOME || process.env.HOMEPATH || process.env.USERPROFILE, 'azure-pipelines-task-lib', '_download');
 
 export class MockTestRunner {
     constructor(testPath: string) {
@@ -128,13 +128,27 @@ export class MockTestRunner {
     private getNodePath(): string {
         const version: number = this.getNodeVersion();
 
+        let downloadVersion: string;
+        switch (version) {
+            case 5:
+                downloadVersion = '5.10.1';
+                break;
+            case 6:
+                downloadVersion = '6.10.3';
+                break;
+            case 10:
+                downloadVersion = '10.15.1';
+                break;
+            default:
+                throw new Error('Invalid node version, must be 5, 6, or 10 (received ' + version + ')');
+        }
+
         // Check if version needed can be found on the path.
         let nodePath: string = shelljs.which('node');
         if (nodePath) {
             try {
                 const output: string = ncp.execSync(nodePath + ' -v').toString().trim();
-                const versionType: string = version + '.x';
-                if (semver.satisfies(output, versionType)) {
+                if (semver.eq(output, downloadVersion)) {
                     return nodePath;
                 }
             }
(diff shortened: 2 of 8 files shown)


node/docs/nodeVersioning.md
```diff
@@ -0,0 +1,20 @@
+# Node versioning
+
+## Agent Node Handler
+
+The agent currently has 2 different node handlers that it can use to execute node tasks: Node 6 and Node 10.
+The handler used depends on the `execution` property specified in the tasks `task.json`.
+If the `execution` property is specified to be `Node`, the task will run on the Node 6 handler, if it is specified to be `Node10` it will run on the Node 10 handler.
+Some legacy agents will use a Node 5 handler if the `execution` is specified to be `Node`, but this is a very small percentage of existing agents.
```
bryanmacfarlane: let's just take this out.  It's so long ago it's probably more of a distraction.  The only agents that are doing this are likely on prem not even dealing with these new tasks.

node/docs/nodeVersioning.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Mock-test Node Handler
+
```
bryanmacfarlane: Add a link to the step by step node testing

node/mock-test.ts
```diff
@@ -1,19 +1,25 @@
...
 import os = require('os');
+import path = require('path');
 import cmdm = require('./taskcommand');
+import semver = require('semver');
 import shelljs = require('shelljs');
+import syncRequest = require('sync-request');
 
 const COMMAND_TAG = '[command]';
 const COMMAND_LENGTH = COMMAND_TAG.length;
+const testDirectory = path.join(__dirname, '_test');
```
bryanmacfarlane: as discussed let's create in home dir, appdata etc...  Not in a folder that gets blown away every time

node/mock-test.ts
```diff
@@ -125,4 +123,195 @@ export class MockTestRunner {
...
         }
     }
-}
+
+    // Returns a path to node.exe with the correct version for this task (based on if its node10 or node)
+    private getNodePath(): string {
+        const version: number = this.getNodeVersion();
+
+        // Check if version needed can be found on the path.
+        let nodePath: string = shelljs.which('node');
```
bryanmacfarlane: let's take out this fuzzy check on current node major version.  we want a crisp run of exactly the versions used by the agent.  this isn't optimizing anything since we're caching in the home dir per above.

damccorm: I left an exact match check because it saves a download on the first run (and thus all CI runs) if its already stored in the path. I'd be comfortable taking it out if you feel strongly though.

damccorm: Taken out per offline discussion

node/mock-test.ts
```diff
@@ -125,4 +123,180 @@ export class MockTestRunner {
...
+            return pathToExe;
+        }
+        else {
+            return this.downloadNode(downloadVersion, downloadDestination);
+        }
+    }
+
+    // Determines the correct version of node to use based on the contents of the task's task.json. Defaults to Node 10.
+    private getNodeVersion(): number {
+        if (process.env['useNodeVersion']) {
```
ericsciple: would it make more sense for discoverability to add optional parameters to the `run` function instead of reading these from env vars?

damccorm: Good call.

node/mock-test.ts
```diff
@@ -125,4 +129,176 @@ export class MockTestRunner {
...
+            case 10:
+                downloadVersion = 'v10.15.1';
+                break;
+            default:
+                throw new Error('Invalid node version, must be 5, 6, or 10 (received ' + version + ')');
+        }
+
+        // Install node in home directory if it isn't already there.
+        const downloadDestination: string = path.join(downloadDirectory, 'node' + version);
+        const pathToExe: string = this.getPathToNodeExe(downloadVersion, downloadDestination);
```
ericsciple: we should check for a marker file so we know the download and extraction finished correctly (not partial file)

node/mock-test.ts
```diff
@@ -125,4 +129,176 @@ export class MockTestRunner {
...
+    }
+
+    // Determines the correct version of node to use based on the contents of the task's task.json. Defaults to Node 10.
+    private getNodeVersion(): number {
+        const taskJsonPath: string = this.getTaskJsonPath();
+        if (!taskJsonPath) {
+            console.warn('Unable to find task.json, defaulting to use Node 10');
+            return 10;
+        }
+        const taskJson: object = require(taskJsonPath);
```
ericsciple: when loading a json file, prefer JSON.parse and fs.readsync over `require`

require caches and this gets really confusing to debug when it goes wrong.

node/mock-test.ts
```diff
@@ -125,4 +129,176 @@ export class MockTestRunner {
...
+
+        const execution: object = taskJson['execution'];
+        let nodeVersion: number = 0;
+        Object.keys(execution).forEach((key: string) => {
+            if (key.toLowerCase() == 'node') {
+                nodeVersion = 6;
+            }
+            else if (key.toLowerCase() == 'node10') {
```
ericsciple: should the order of the if conditions be switched? in other words, prefer 10 over 6?

ericsciple: Another thing we could do here that would be helpful, is check for a min agent version in the task.json. Node 6 wasn't added to the agent until 2.117.0, per [this document](https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/node/docs/minagent.md#task-execution-handler)

node/mock-test.ts
```diff
@@ -125,4 +129,176 @@ export class MockTestRunner {
...
+                return path.join(downloadDestination, 'node-' + nodeVersion + '-linux-x64', 'bin', 'node');
+            case 'win32':
+                this.downloadFile(nodeUrl + '/' + nodeVersion + '/win-x64/node.exe', downloadDestination, 'node.exe');
+                this.downloadFile(nodeUrl + '/' + nodeVersion + '/win-x64/node.lib', downloadDestination, 'node.lib');
+                return path.join(downloadDestination, 'node.exe')
+        }
+    }
+
+    //Downloads file to the downloadDestination, making any necessary folders along the way.
+    private downloadFile(url: string, downloadDestination: string, fileName: string): void {
```
ericsciple: `// Downloads` (missingaspace)

node/mock-test.ts
```diff
@@ -125,4 +129,176 @@ export class MockTestRunner {
...
+        if (!url) {
+            throw new Error('Parameter "url" must be set.');
+        }
+        if (!downloadDestination) {
+            throw new Error('Parameter "downloadDestination" must be set.');
+        }
+        console.log('Downloading file:', url);
+        shelljs.mkdir('-p', downloadDestination);
```
ericsciple: should we just use fs? move away from npm install warnings

node/mock-test.ts
```diff
@@ -125,4 +129,176 @@ export class MockTestRunner {
...
+        if (platform != 'darwin' && platform != 'linux' && platform != 'win32') {
+            throw new Error('Unexpected platform: ' + platform);
+        }
+        return platform;
+    }
+
+    private cd(dir: string): void {
+        var cwd: string = process.cwd();
+        if (cwd != dir) {
+            shelljs.cd(dir);
```
ericsciple: consider process.chdir

node/mock-test.ts
```diff
@@ -125,4 +129,176 @@ export class MockTestRunner {
...
+        if (!url) {
+            throw new Error('Parameter "url" must be set.');
+        }
+        if (!downloadDestination) {
+            throw new Error('Parameter "downloadDestination" must be set.');
+        }
+        console.log('Downloading file:', url);
+        shelljs.mkdir('-p', downloadDestination);
+        const result: any = syncRequest('GET', url);
+        fs.writeFileSync(path.join(downloadDestination, fileName), result.getBody());
```
ericsciple: is this how we do the download in other places? i'm guessing this throws on unsuccessful http response codes? does this handle redirects?

we have similar code in the task tool lib and either other places in this repo or the tasks repo or the agent repo

node/mock-test.ts
```diff
@@ -1,19 +1,27 @@
...
 import os = require('os');
+import path = require('path');
 import cmdm = require('./taskcommand');
+import semver = require('semver');
 import shelljs = require('shelljs');
+import syncRequest = require('sync-request');
 
 const COMMAND_TAG = '[command]';
 const COMMAND_LENGTH = COMMAND_TAG.length;
+const downloadDirectory = path.join(process.env.HOME || process.env.HOMEPATH || process.env.USERPROFILE, 'azure-pipelines-task-lib', '_download');
```
ericsciple: where does this go on windows? should we use appdatalocal?

node/mock-test.ts
```diff
@@ -125,4 +129,176 @@ export class MockTestRunner {
...
+        }
+
+        // Install node in home directory if it isn't already there.
+        const downloadDestination: string = path.join(downloadDirectory, 'node' + version);
+        const pathToExe: string = this.getPathToNodeExe(downloadVersion, downloadDestination);
+        if (pathToExe) {
+            return pathToExe;
+        }
+        else {
+            return this.downloadNode(downloadVersion, downloadDestination);
```
ericsciple: do we need to add node to the PATH?

node/docs/nodeVersioning.md
```diff
@@ -0,0 +1,20 @@
+# Node versioning
```
ericsciple: i left some comments of things to think through. otherwise looks good to me

node/mock-test.ts
damccorm: We could, I think this is probably fine. [HomePath](https://environmentvariables.org/HomePath) should resolve this, if not [UserProfile](http://environmentvariables.org/UserProfile) should, either way should point to Users\{username}

node/mock-test.ts
damccorm: No, as we discussed offline this mimics the agent behavior.

node/mock-test.ts
damccorm: Decided to leave out node 5 check per our offline conversation, do now prefer node 10 though.

node/mock-test.ts
damccorm: I think I like leaving this for now because `fs.mkdirSync` doesn't have a recursive option in Node<10. I can make that, but I don't feel a need to do that until we decide we're completely getting rid of the dependency.

node/docs/nodeVersioning.md
damccorm: I responded to everything, could probably use another review before merging.

node/mock-test.ts
```diff
@@ -123,4 +128,186 @@ export class MockTestRunner {
...
+            if (fs.existsSync(taskJsonPath)) {
+                return taskJsonPath;
+            }
+            newPath = path.join(curPath, '..');
+        }
+        return '';
+    }
+
+    // Downloads the specified node version to the download destination. Returns a path to node.exe
+    private downloadNode(nodeVersion: string, downloadDestination: string): string {
```
ericsciple: should you rmdir the downloadDestination first?

damccorm: Right now I remove files in downloadFile, I think I agree that I like this approach more though.

damccorm: Updated accordingly.

node/mock-test.ts
ericsciple: I'm fine either way.

Note, mkdirP is in task lib. do we have access to that function here?

node/docs/nodeVersioning.md
ericsciple: I left a couple small comments. Looks good to me. Nice work.

node/mock-test.ts
damccorm: Not without import `task.ts` which would probably be fine, but I'm a little more inclined to leave this for now and handle it (along with the rest of shelljs) in a future PR.

PR #505, opened 2019-03-08
Updated vault to encode secrets using cipheriv

### LINKED ISSUE #380: Include support for Node 8.7+

damccorm:
Branch got messed up because of rebrand and me deleting my fork 🤦‍♂️ and the easiest thing to do was just create a new PR (moved from #416)

Fixes #380

### DIFF
--- a/node/docs/releases.md
+++ b/node/docs/releases.md
@@ -1,13 +1,14 @@
 # AZURE-PIPELINES-TASK-LIB RELEASES
 
 ## 2.8.0
- * Updated to TypeScript 3.0
+ * Updated to TypeScript 3.0.
  * Fixed `which` so that it finds executables that can be executed by the current process, not just those that are executable by everyone.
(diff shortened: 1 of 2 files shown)


damccorm:
From #416, TODOs are 
- [ ] move the duplicate logic into `getKey`
- [ ] make sure it works on node 5
- [ ] talk to CJ about sha256 vs 512 and difference between the hash approach and scrypt.

Additionally:
- [ ] Tests are now failing, figure out why.

PR #511, opened 2019-03-20
OSX/Linux platform - handle double quotes while logging command

ajinkya599:
Double quotes character present in the arguments are printed without the escape character (\\). Adding it explicitly to the cmd string.

Ex. A command mycmd running with an argument myarg (myarg = 'my\\"arg'),
is logged as: mycmd my"arg
but its actually executed as: mycmd my\\"arg

### DIFF
--- a/node/test/toolrunnertests.ts
+++ b/node/test/toolrunnertests.ts
@@ -1254,6 +1254,43 @@ describe('Toolrunner Tests', function () {
                     done(err);
                 });
         });
+
+        // ----------------------------------------
+        // double quotes in arg (OSX/Linux)
+        // ----------------------------------------
+
+        it('exec with double quotes in args (OSX/Linux)', function (done) {
+            this.timeout(10000);
+            let bashPath = tl.which('bash');
+            let bash = tl.tool(bashPath)
+                .arg('-c')
+                .arg('echo $0')
+                .arg('hello"world"')
+            let outStream = testutil.createStringStream();
+            let options = <trm.IExecOptions>{ outStream: <stream.Writable>outStream, windowsVerbatimArguments: true };
+            let output = '';
+            bash.on('stdout', (data) => {
+                output += data.toString();
+            });
+            bash.exec(options)
+                .then(function (code) {
+                    assert.equal(code, 0, 'return code should be 0');
+                    // validate the [command] header
+                    assert.equal(
+                        outStream.getContents().split(os.EOL)[0],
+                        '[command]' + bashPath + ' -c echo $0 hello\\"world\\"',
+                        '[command] header should match');
+                    // validate stdout
+                    assert.equal(
+                        output.trim(),
+                        'hello"world"',
+                        'stdout should match');
+                    done();
+                })
+                .fail(function (err) {
+                    done(err);
+                });
+        });
     }
     else { // process.platform == 'win32'
 
--- a/node/toolrunner.ts
+++ b/node/toolrunner.ts
@@ -170,6 +170,10 @@ export class ToolRunner extends events.EventEmitter {
             // on Unix, execvp() takes an arg array.
             cmd += toolPath;
             args.forEach((a: string): void => {
+                // Prefix all occurences of '"' character with '\' character explicitly,
+                // so that the arguments containing double quotes are printed correctly
+                // in the logs.
+                a = a.replace(/\"/g, "\\\"");
                 cmd += ` ${a}`;
             });
         }
--- a/node/toolrunner.ts
+++ b/node/toolrunner.ts
@@ -170,9 +170,15 @@ export class ToolRunner extends events.EventEmitter {
             // on Unix, execvp() takes an arg array.
             cmd += toolPath;
             args.forEach((a: string): void => {
-                // Prefix all occurences of '"' character with '\' character explicitly,
+                // Prefix all occurences of " character with \ character explicitly,
                 // so that the arguments containing double quotes are printed correctly
                 // in the logs.
+                // Without this, secret masking does not work when the original input
+                // contains escaped double quotes character.
+                // Example: An argument entered as example\"string in the UI, will have
+                // value example"string. While passing the argument to the command, it
+                // is passed as example\"string, where as in the logs, it appears as
+                // example"string.
                 a = a.replace(/\"/g, "\\\"");
                 cmd += ` ${a}`;
             });


node/toolrunner.ts
```diff
@@ -170,6 +170,10 @@ export class ToolRunner extends events.EventEmitter {
             // on Unix, execvp() takes an arg array.
             cmd += toolPath;
             args.forEach((a: string): void => {
+                // Prefix all occurences of '"' character with '\' character explicitly,
+                // so that the arguments containing double quotes are printed correctly
+                // in the logs.
+                a = a.replace(/\"/g, "\\\"");
```
ds-ms: What would happen if `"` is already escaped? 
Like `example\"string` would become `example\\"string` and would end up escaping `\` instead of `"` in that case?

damccorm:
I think I'm not 100% clear on what the desired behavior is here. Here's my current understanding

| Argument value | Current Output | Desired output |
|---|---|---|
| `a"b` | `a\"b` | `a\"b` |
| `a\"b` | `a\\"b` | `a\"b` (or should this be `a\\\"b`?) |

@ajinkya599, @ds-ms is this accurate? @ericsciple any thoughts?

ajinkya599:
> 
> 
> I think I'm not 100% clear on what the desired behavior is here. Here's my current understanding
> Argument value 	Current Output 	Desired output
> `a"b` 	`a\"b` 	`a\"b`
> `a\"b` 	`a\\"b` 	`a\"b` (or should this be `a\\\"b`?)
> 
> @ajinkya599, @ds-ms is this accurate? @ericsciple any thoughts?

I am coming from [#9633](https://github.com/Microsoft/azure-pipelines-tasks/issues/9633), where (as I mentioned in the [comment](https://github.com/Microsoft/azure-pipelines-tasks/issues/9633#issuecomment-474839460)), the desired behavior can be understood as:

| Argument value | Current Output (Windows) | Current Output (OSX/Linux) | Desired Output (All platforms) |
| :---: | :---: | :---: | :---: |
| `a"b` | `a\"b` | `a"b` | `a\"b` |

i.e. the output in the logs should be in parity with the actual command getting executed.

NOTE:

- What I mean by argument value here is the value contained in `a` and not the original input. i.e.,
`a == 'a"b'` or `a == "a\"b"`. (This actually happens when the original input is `a\"b`)
- Here, the argument is created using Release variable having value `a\"b`. When this argument reaches this code its value is `a"b`. It is passed to the command as `a\"b` which is desired, but logged as `a"b` for OSX/Linux which is not what's desired (for Windows its logged as `a\"b`, as desired).
- Many other cases (with `\`) are handled [here](https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/node/toolrunner.ts#L431), but far as [#9633](https://github.com/Microsoft/azure-pipelines-tasks/issues/9633) is concerned, fixing the above case is sufficient. (It may make sense to fix the cases for `\` as well).

node/toolrunner.ts
ajinkya599: Discussed offline. Adding comment for reader's clarity:

An escaped double quote is actually what we are dealing with here.
`a` having value `"example\"string"` (` = 'example"string'`), happens when actual input is escaped, i.e. the argument is defined as `example\"string` in the UI input. When printed to the logs the output is `example"string` for OSX/Linux and `example\"string` for Windows.

While, when passed to the command, it is passed as `example\"string` for both Windows as well as OSX/Linux.

In case the actual input is not escaped (`example"string`), the value of `a` ends up to be `examplestring` and this code does nothing to it.

damccorm:
> Many other cases (with \) are handled here

That's only called on the Windows case, I think we can limit our discussion to the Unix case since that's where the issue arises.

> but far as #9633 is concerned, fixing the above case is sufficient. (It may make sense to fix the cases for \ as well).

Sure, and I'm ok with us making incremental progress. I don't however, want to introduce a behavior that could break someone or cause further issues.

To be clear, it seems to me like the way you are handling the `a"b` case you talk about in your comment correctly. It doesn't, however, seem like you're handling the `a\"b` case correctly though, and you're changing behavior.

I'm ok changing behavior since we're in the midst of a major version upgrade, but we need to make sure its the right change.

node/toolrunner.ts
damccorm: Ok, this was helpful. I think I'm more convinced this is the right change. Continuing general conversation below.

damccorm:
From above, I'm more convinced this is the right thing to be doing. I think I'd be in favor of merging if:
1) We add tests that currently fail but don't after this fix.
2) We add a comment explaining why we're doing this in the code.

damccorm:
@ajinkya599 do you think you'll have time to take this PR the rest of the way?

damccorm:
@ajinkya599 any update on this? I'm happy to take this the rest of the way if you don't have the bandwidth, but I'd like to get the major version out of preview soon, and would like to get this in the new major version.

ajinkya599:
Hey @damccorm, sorry for the delayed response as I was caught up with Build deliverables and missed replying on this thread. I am on vacation right now and will be back on Sunday. 
I should be able to work on this PR post next week. Does that sound fine ?

damccorm:
Sure, that's fine. We ended up deciding not to major version at this point, so this will just go out as a minor patch - I think that's fine since this is just changing logging behavior.

moswald:
@damccorm what's the status on this? I am reviewing our issues list and this PR is referenced by #9633. It hasn't seen any action since mid-March, so I'm trying to bring it back on someone's radar.

damccorm:
@ajinkya599 any update? Would love it if we could take this the rest of the way, I think its pretty close

ajinkya599:
Added the test and comments. @damccorm - Please review the changes.
Do the tests not run as a part of the PR ?

damccorm:
> Do the tests not run as a part of the PR ?

Looks like we had it turned off for forked builds for some reason, changed it so that it should run from now on.

damccorm:
Looks like your test is failing (ignore the windows failure, its unrelated/flaky)

stephenmichaelf:
This should be reviewed and approved by @ericsciple before merging.

ajinkya599:
Thanks @damccorm for enabling the tests. Updated the test and now it is passing.
@ericsciple - Can you please take a look at this PR ?

ericsciple:
Since the problem is the `line` function, I don't think changing a downstream function makes sense. The proposed change has wider consequences.

Also if recall correctly, this is only an issue for an old-major version of one specific task.

Have you considered copying the `line` function into your task instead? Then you can override the behavior as desired?

The `line` function is mutating the data before it calls `.arg(...)`.

node/test/toolrunnertests.ts
```diff
@@ -1254,6 +1254,43 @@ describe('Toolrunner Tests', function () {
...
+        // double quotes in arg (OSX/Linux)
+        // ----------------------------------------
+
+        it('exec with double quotes in args (OSX/Linux)', function (done) {
+            this.timeout(10000);
+            let bashPath = tl.which('bash');
+            let bash = tl.tool(bashPath)
+                .arg('-c')
+                .arg('echo $0')
+                .arg('hello"world"')
```
zachariahcox: this does not appear to be testing the issue when the secret was literally: 
hello\"secret

zachariahcox:
I tend to agree with eric.
What is the argument for making this change here and not just editing the kub task? It does seem to be literally mutating the inputs, which is wrong.

ajinkya599:
> What is the argument for making this change here and not just editing the kub task?
>
I don't see what change can be done in kube task to make this work.
IMO, everything is fine end-to-end except for the logging done in the `_getCommandString` function for `Linux/OSX`.


Can you share what can be changed in the kube task ?

> It does seem to be literally mutating the inputs, which is wrong.
>

It is nodejs behavior that when it is passed an argument containing `"` character, it escapes it using `\` character.
That is, when you tell node to spawn a child process and execute `sometool` with argument `hello"world`, its execution (the end result) is equivalent to executing `sometool hello\"world` on the cli.

Also, if mutating (to make it equivalent to the actual behavior) the `input` while **logging** is wrong, why is it done in Windows (regular) case ?
[_getCommandString](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/toolrunner.ts#L169) -> [_windowsQuoteCmdArg](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/toolrunner.ts#L295) -> [_uv_quote_cmd_arg](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/toolrunner.ts#L439)

node/test/toolrunnertests.ts
ajinkya599: The test is for the case when the secret is literally `hello\"secret`. (When the secret is `hello"secret`, the `"` gets eaten-up somewhere in between, and the value here is `helloworld`). This discussion is there in the conversation.

damccorm:
> I don't see what change can be done in kube task to make this work.

You definitely _could_ fix this in the kube task by just making sure we mark arguments like this as secrets (whether they have " or not in them) - e.g. add an extra mask for what gets printed out by this function.

-------------------------------------------------------------------------------------

As I revisit this, I don't think we should do this. I think I have a couple reasons:

1) I think it breaks the mainline secrets use case. Right now, if I set a secret `Foo\"bar` in node, the command output ends up being `##vso[task.setSecret variable=Foo"bar;isSecret=true]myVal` - so if I were to then print out the arg value as you're suggesting in this PR (`Foo\"bar`), it would not get masked.
2) This ends up being inconsistent with all of our other escaping. Even with your change, `Foo\tbar` would get printed out as `Foo   bar`, `Foo\'bar` would get printed out as `Foo'bar`, and even `Foo\\bar` would get printed out as `Foo\bar`. If we do end up taking this, we probably want to do it for all of our escaping, which feels much less palatable.

At the end of the day, we don't know how the tool that is using the args is going to process them, so I don't think we should mutate them.

ajinkya599:
>You definitely could fix this in the kube task by just making sure we mark arguments like this as secrets (whether they have " or not in them) - e.g. add an extra mask for what gets printed out by this function.
>
Are you suggesting that we parse all the arguments given to the task and explicitly mask each one of them after removing the escape character (`\`) ? That looks hacky and not something that we would want to do.

>1. I think it breaks the mainline secrets use case. Right now, if I set a secret Foo\"bar in node, the command output ends up being ##vso[task.setSecret variable=Foo"bar;isSecret=true]myVal - so if I were to then print out the arg value as you're suggesting in this PR (Foo\"bar), it would not get masked.
>
(Doesn't this concern hold true for Windows case ? )
Anyways, I tried `setVariable("myvar", "my\"secret", true)` and it seems to mask both `my\"secret` and `my"secret` in the logs. (I printed these literals using cli task, and both got masked.)
So, if we are printing out the arg value as per this PR(`my\"secret`), it would get masked as well.

>2. This ends up being inconsistent with all of our other escaping. Even with your change, Foo\tbar would get printed out as Foo bar, Foo\'bar would get printed out as Foo'bar, and even Foo\\bar would get printed out as Foo\bar. If we do end up taking this, we probably want to do it for all of our escaping, which feels much less palatable.
>
Agree that there are other escaping cases that should be handled as well. But as we [discussed](https://github.com/microsoft/azure-pipelines-task-lib/pull/511#issuecomment-480941268) earlier, the scope of this PR is for the `\"` case. Incremental progress can be made on this. IMO, the other cases should also be handled (separately), as this is a potential security issue for any task that is consuming task-lib for tool execution.

>At the end of the day, we don't know how the tool that is using the args is going to process them, so I don't think we should mutate them.
>
We are not mutating the input args to the tool. We are just amending the logs to bring it in parity with the node behavior. (We are dealing with node behavior to escape the `"` character by default. The tool receives escaped argument, but what gets logged is un-escaped, which is incorrect.)

Also, I am unable to understand what is the difference between Windows and Unix/OSX case here. Didn't most of these concerns exist for Windows case as well ?

damccorm:
> Anyways, I tried setVariable("myvar", "my\"secret", true) and it seems to mask both my\"secret and my"secret in the logs. (I printed these literals using cli task, and both got masked.)

Hm, that is true - forgot we were a little intentionally aggressive with our secret masking. So you're right, that's probably not a concern.

> Agree that there are other escaping cases that should be handled as well. But as we discussed earlier, the scope of this PR is for the \" case. IMO, the other cases should also be handled (separately), as this is a potential security issue for any task that is consuming task-lib for tool execution.

I think I'm somewhat ok with incremental progress, but it feels very inconsistent to do this for `"` and not `'`. We're basically saying that this is a security hole (which I'm still not totally convinced of), but we're only fixing it for one of the mainline use cases. I think I'm less worried about the other use cases (though couldn't those still be issues for your task?). The issue here is that we don't mask `my\'secret` if you call `setVariable("myvar", "my\'secret", true)`, we just mask what actually gets printed to stdout (`my'secret` in this case). So this still breaks the mainline secret use case.

> We are not mutating the input args to the tool. We are just amending the logs to bring it in parity with the node behavior. (We are dealing with node behavior to escape the " character by default. The tool receives escaped argument, but what gets logged is un-escaped, which is incorrect.)

You're right - statement retracted 😃 I got a little lost in the weeds.

> Also, I am unable to understand what is the difference between Windows and Unix/OSX case here. Didn't most of these concerns exist for Windows case as well ?

Not totally sure what you're getting at here. The Windows escaping you linked to above is specifically for appending arguments because of how windows parses additional args (and that's what's executed not just what's logged). We don't need that for linux because the way args are appended are different - see the comment right above your changes and the preceding comments:
> on Unix, execvp() takes an arg array.

----------------------------------------------------------------------------------------------------------------

I think I'm back to thinking this is an ok idea, but I don't really want to do it until we find a more extensible solution that doesn't leave us in a weird inconsistent state.

I wonder if this should just be an masking side change, where any time we add something with an escapable character, we also add the same string with the unescaped versions. So if you pass in `FOO\BAR"ABC'DEF` to be masked, we'd also mask `FOO\\BAR\"ABC\'DEF`.

Honestly, I'm still not convinced that the right solution here isn't just for your task to add an extra mask for what gets printed out by this function.

ajinkya599:
>
>I think I'm back to thinking this is an ok idea, but I don't really want to do it until we find a more extensible solution that doesn't leave us in a weird inconsistent state.
>

I think I agree with you on finding an extensible solution. For now, we have suggested a [work-around](https://github.com/microsoft/azure-pipelines-tasks/issues/9633#issuecomment-568755757) (Using Powershell script, avoiding tool execution through task-lib).

>
>I wonder if this should just be an masking side change, where any time we add something with an escapable character, we also add the same string with the unescaped versions. So if you pass in FOO\BAR"ABC'DEF to be masked, we'd also mask FOO\\BAR\"ABC\'DEF.
>

I like this idea! 😊 More aggressive masking can protect other unknown potential leakages as well.

I would leave it upto you guys to decide the right thing and take this forward. Please let me know if you need any inputs from my side.

I am leaving the original issue ([#9633](https://github.com/microsoft/azure-pipelines-tasks/issues/9633)) open. Should I abandon this PR ?

damccorm:
> Should I abandon this PR ?

I think so

PR #518, opened 2019-03-26
Remove shelljs dependency

damccorm:


node/mock-test.ts
```diff
@@ -211,7 +211,7 @@ export class MockTestRunner {
 
     // Downloads the specified node version to the download destination. Returns a path to node.exe
     private downloadNode(nodeVersion: string, downloadDestination: string): string {
-        shelljs.rm('-rf', downloadDestination);
+        task.rmRF(downloadDestination);
```
stephenmichaelf: Do we want to have these hard coded to functions (rmRf, mkdirP) or keep the syntax similar to what it was?

node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "3.0.2-preview",
+  "version": "3.0.3-preview",
```
stephenmichaelf: Do we need preview description?

node/package.json
```diff
@@ -29,11 +29,11 @@
     "mockery": "^1.7.0",
     "q": "^1.1.2",
     "semver": "^5.1.0",
-    "shelljs": "^0.3.0",
     "uuid": "^3.0.1"
   },
   "devDependencies": {
     "mocha": "5.2.0",
+    "shelljs": "^0.3.0",
```
stephenmichaelf: Can we get rid of this?

node/task.ts
```diff
@@ -1,5 +1,5 @@
 import Q = require('q');
-import shell = require('shelljs');
+import child_process = require('child_process');
```
stephenmichaelf: import childProcess?

node/task.ts
```diff
@@ -636,8 +637,10 @@ export function pushd(path: string): void {
  * @returns   void
  */
 export function popd(): void {
-    shell.popd();
-    _checkShell('popd');
+    const path: string = dirStack.pop() || "";
+    if (path != "") {
```
stephenmichaelf: !==

node/task.ts
```diff
@@ -725,36 +728,104 @@ export function resolve(...pathSegments: any[]): string {
...
-export function ls(options: string, paths: string[]): string[] {
-    if (options) {
-        return shell.ls(options, paths);
-    } else {
-        return shell.ls(paths);
-    }
+export function ls(paths: string[], recursive: boolean = false, allFiles: boolean = false): string[] {
+    let filePaths: string[] = [];
+    paths.forEach(path => {
+        let commandString = '';
```
stephenmichaelf: let command: string?

node/task.ts
```diff
@@ -725,36 +728,104 @@ export function resolve(...pathSegments: any[]): string {
...
+        else {
+            commandString = 'ls ' + path;
+            if (recursive) {
+                commandString += ' -R';
+            }
+            if (allFiles) {
+                commandString += ' -A';
+            }
+        }
+        child_process.execSync(commandString).toString().split(os.EOL).forEach((file: string) => {
```
stephenmichaelf: Can we split this out, confusing in one line.

node/task.ts
```diff
@@ -725,36 +728,104 @@ export function resolve(...pathSegments: any[]): string {
...
+                if (fs.existsSync(dest)) {
+                    if (force) {
+                        fs.unlinkSync(dest);
+                    }
+                    else {
+                        // If file exists and we're not overwriting, just return. 
+                        return;
+                    }
+                }
+                let commandString = 'echo F | xcopy ' + source + ' ' + dest;
```
stephenmichaelf: const command: string

node/task.ts
```diff
@@ -725,36 +728,104 @@ export function resolve(...pathSegments: any[]): string {
...
-        shell.cp(options, source, dest);
+export function cp(source: string, dest: string, recursive: boolean = false, force: boolean = false, continueOnError?: boolean): void {
+    try {
+        if (!fs.existsSync(source)) {
+            throw new Error(loc('LIB_OperationFailed', source + ' doesnt exist'));
+        }
+        if (getPlatform() == Platform.Windows) {
+            if (fs.statSync(source).isDirectory()) {
+                mkdirP(dest);
+                let commandString = 'robocopy ' + source + ' ' + dest;
```
stephenmichaelf: Missing type.

node/task.ts
```diff
@@ -725,36 +728,104 @@ export function resolve(...pathSegments: any[]): string {
...
+                        // If file exists and we're not overwriting, just return. 
+                        return;
+                    }
+                }
+                let commandString = 'echo F | xcopy ' + source + ' ' + dest;
+                child_process.execSync(commandString);
+            }
+        }
+        else {
+            let commandString = 'cp ';
```
stephenmichaelf: command: string

node/task.ts
```diff
@@ -725,36 +728,104 @@ export function resolve(...pathSegments: any[]): string {
...
+            if (recursive) {
+                commandString += '-R ';
```
stephenmichaelf: Should we be consistent with above? Add space before param.

node/task.ts
```diff
@@ -765,15 +836,55 @@ export function cp(source: string, dest: string, options?: string, continueOnErr
  * @param     options    string -f or -n for force and no clobber 
  * @param     continueOnError optional. whether to continue on error
  */
-export function mv(source: string, dest: string, options?: string, continueOnError?: boolean): void {
-    if (options) {
-        shell.mv(options, source, dest);
+export function mv(source: string, dest: string, force = false, continueOnError?: boolean): void {
```
stephenmichaelf: Add type to force.

node/task.ts
```diff
@@ -765,15 +836,55 @@ export function cp(source: string, dest: string, options?: string, continueOnErr
...
+                
+                if (force) {
+                    commandString += ' /is /it';
+                }
+
+                child_process.execSync(commandString);
+            }
+            else {
+                // Copy individual file over
+                if (fs.existsSync(dest)) {
```
stephenmichaelf: Move this up to the top?

node/task.ts
```diff
@@ -765,15 +836,55 @@ export function cp(source: string, dest: string, options?: string, continueOnErr
...
-        shell.mv(options, source, dest);
+export function mv(source: string, dest: string, force = false, continueOnError?: boolean): void {
+    try {
+        if (!fs.existsSync(source)) {
+            throw new Error(loc('LIB_OperationFailed', source + ' doesnt exist'));
+        }
+        if (getPlatform() == Platform.Windows) {
+            if (fs.statSync(source).isDirectory()) {
+                mkdirP(dest);
+                let commandString = 'robocopy ' + source + ' ' + dest + ' /move';
```
stephenmichaelf: Add type.

node/task.ts
```diff
@@ -765,15 +836,55 @@ export function cp(source: string, dest: string, options?: string, continueOnErr
...
+                if (fs.existsSync(dest)) {
+                    if (force) {
+                        fs.unlinkSync(dest);
+                    }
+                    else {
+                        // If file exists and we're not overwriting, just return.
+                        return;
+                    }
+                }
+                let commandString = 'echo F | xcopy ' + source + ' ' + dest;
```
stephenmichaelf: Missing type.

node/task.ts
```diff
@@ -765,15 +836,55 @@ export function cp(source: string, dest: string, options?: string, continueOnErr
...
+                child_process.execSync(commandString);
+                fs.unlinkSync(source);
+            }
+        }
+        else {
+            let commandString = 'mv ';
```
stephenmichaelf: Missing type.

node/task.ts
```diff
@@ -765,15 +836,55 @@ export function cp(source: string, dest: string, options?: string, continueOnErr
...
+            if (force) {
+                commandString += '-f ';
```
stephenmichaelf: Same comment about spacing.

node/task.ts
```diff
@@ -1145,45 +1256,44 @@ function _legacyFindFiles_getMatchingItems(
...
-    }
-    catch (err) {
-        // if you try to delete a file that doesn't exist, desired result is achieved
-        // other errors are valid
-        if (err.code == 'ENOENT') {
-            return;
+export function rmRF(inputPath: string): void {
+    debug('rm -rf ' + inputPath);
+    if (fs.existsSync(inputPath)) {
+        const inputStats: fs.Stats = fs.lstatSync(inputPath);
```
stephenmichaelf: Can this return null?

node/task.ts
```diff
@@ -1145,45 +1256,44 @@ function _legacyFindFiles_getMatchingItems(
...
+        if (inputStats.isDirectory()) {
+            fs.readdirSync(inputPath).forEach(function(file, index){
```
stephenmichaelf: Nit: space before open brace

stephenmichaelf:
Were you able to do a review of all the tests to make sure we don't have any gaps?

node/mock-test.ts
damccorm: I'm not sure I understand what you're asking here - are you concerned about us using functions from task here? Or is it something else?

node/package.json
damccorm: We still use it in some of our build process. I think that would be the next step, but that feels secondary to the primary goal of getting rid of any functional impact from shelljs (and is somewhat orthogonal).

To be clear, I want to get rid of it, just also want to keep PRs as focused/manageable as possible.

node/task.ts
damccorm: Not if `inputPath` exists (which we've already checked).

node/task.ts
damccorm: Sure

node/package.json
stephenmichaelf: Makes sense, can we create an issue to track it.

node/package.json
stephenmichaelf: Oh I meant the other way :) we have 3.0.3-preview, is there any reason we don't have 3.0.3?

stephenmichaelf: I guess since we are doing new major it's probably fine.

node/mock-test.ts
stephenmichaelf: Ignore this comment, I misunderstood the surface area we expose.

node/package.json
damccorm: Added #523 to track this

damccorm:
@stephenmichaelf am I good to merge this?

stephenmichaelf:
Let's wait for @ericsciple to review. He has more context on the code.

bryanmacfarlane:
I want review tomorrow as well.  Please hold

node/task.ts
```diff
@@ -614,20 +597,38 @@ export const checkPath = im._checkPath;
...
+        let errMsg: string = '';
+        if (!fs.existsSync(path)) {
+            errMsg = loc('LIB_OperationFailed', 'cd ' + path, 'No such file or directory');
+        }
+        else if (!fs.statSync(path).isDirectory()) {
+            errMsg = loc('LIB_OperationFailed', 'cd ' + path, 'Not a directory');
+        }
+
+        if (!errMsg) {
+            process.chdir(path);
```
stephenmichaelf: Can this throw?

node/task.ts
```diff
@@ -725,36 +728,105 @@ export function resolve(...pathSegments: any[]): string {
...
  */
-export function ls(options: string, paths: string[]): string[] {
-    if (options) {
-        return shell.ls(options, paths);
-    } else {
-        return shell.ls(paths);
-    }
+export function ls(paths: string[], recursive: boolean = false, allFiles: boolean = false): string[] {
+    let filePaths: string[] = [];
+    paths.forEach(path => {
```
stephenmichaelf: Check for null on paths

node/task.ts
```diff
@@ -725,36 +728,105 @@ export function resolve(...pathSegments: any[]): string {
...
+ * @param     recursive  Copy recursively
+ * @param     force      Force copy
  * @param     continueOnError optional. whether to continue on error
  */
-export function cp(source: string, dest: string, options?: string, continueOnError?: boolean): void {
-    if (options) {
-        shell.cp(options, source, dest);
+export function cp(source: string, dest: string, recursive: boolean = false, force: boolean = false, continueOnError?: boolean): void {
+    try {
+        if (!fs.existsSync(source)) {
```
stephenmichaelf: Check dest too?

stephenmichaelf: What if dest exists and we are copying a directory.

node/task.ts
```diff
@@ -725,36 +728,105 @@ export function resolve(...pathSegments: any[]): string {
...
+                command += ' -f';
+            }
+            command += ' ' + source + ' ' + dest;
+            childProcess.execSync(command);
+        }
     }
-    else {
-        shell.cp(source, dest);
+    catch (err) {
+        if (!continueOnError) {
```
stephenmichaelf: Should we still write to the log that there was an error?

node/task.ts
```diff
@@ -765,15 +837,55 @@ export function cp(source: string, dest: string, options?: string, continueOnErr
...
-    if (options) {
-        shell.mv(options, source, dest);
+export function mv(source: string, dest: string, force: boolean = false, continueOnError?: boolean): void {
+    try {
+        if (!fs.existsSync(source)) {
+            throw new Error(loc('LIB_OperationFailed', source + ' doesnt exist'));
+        }
+        if (getPlatform() == Platform.Windows) {
+            if (fs.statSync(source).isDirectory()) {
+                mkdirP(dest);
```
stephenmichaelf: What if it already exists?

damccorm: That's fine, mkdirP creates the folder/all parents only if it/they don't exist

node/task.ts
damccorm: Its fine if dest does or doesn't exist. That's the significance of `force` - it determines if we'll overwrite the file or not. Robocopy handles that for directories no problem, it doesn't do it for files correctly though which is why I have the logical split.

node/task.ts
damccorm: I think we actually are ok leaving this out. If path is null it will call `dir  /b [/s] [/a]` or `ls [-R] [-A]` which will return directory info for the current directory. That seems like the behavior we'd want (its equivalent to passing in a space or period)

node/task.ts
damccorm: Only in exceptional situations (where I think we'd want to throw). The directory is guaranteed to exist/actually be a directory so if `chdir` is throwing that indicates something weird is happening.

node/task.ts
damccorm: Yeah (here and other places). Good call, I'll add that

stephenmichaelf:
Bump @bryanmacfarlane @ericsciple

bryanmacfarlane:
I want to revisit what we are doing with a "new major version" in light of recent plans.  I still want to do it and this is part of that work.  But let's step back and discuss the scope of the new major version again.  Chat offline.

damccorm:
Going to double down on test coverage and make sure to fix the rmrf issue. Will update when this is ready for review again.

damccorm:
@bryanmacfarlane @ericsciple @stephenmichaelf I think this is ready for review.

node/task.ts
```diff
@@ -614,20 +597,38 @@ export const checkPath = im._checkPath;
  */
 export function cd(path: string): void {
     if (path) {
-        shell.cd(path);
-        _checkShell('cd');
+        let errMsg: string = '';
+        if (!fs.existsSync(path)) {
```
ericsciple: iirc the node pattern is do not call `existSync`, just `statSync` and catch ENOENT

node/task.ts
```diff
@@ -614,20 +597,38 @@ export const checkPath = im._checkPath;
...
+            errMsg = loc('LIB_OperationFailed', 'cd ' + path, 'No such file or directory');
+        }
+        else if (!fs.statSync(path).isDirectory()) {
+            errMsg = loc('LIB_OperationFailed', 'cd ' + path, 'Not a directory');
```
ericsciple: `Not a directory` is not localized

ericsciple: consider: why present the error message to the user as a command line instruction? The `cd` command is not actually being invoked.

node/task.ts
ericsciple: also i think the statSync function is in the task lib, it may have a localized error already for the ENOENT case

node/task.ts
```diff
@@ -766,14 +872,73 @@ export function cp(source: string, dest: string, options?: string, continueOnErr
...
+        options = options.toLowerCase();
     }
+    try {
+        if (!fs.existsSync(source)) {
+            throw new Error(loc('LIB_OperationFailed', source + ' doesnt exist'));
+        }
+        if (getPlatform() == Platform.Windows) {
+            if (fs.statSync(source).isDirectory()) {
+                mkdirP(dest);
+                let command: string = 'robocopy ' + source + ' ' + dest + ' /move';
```
ericsciple: - robocopy retries 1 million times by default (too much)
- we should use tool runner
  - it handles quoting
  - it roots the tool (otherwise additional quoting problems on windows)
  - it avoids execSync (buffers output until tool completes - i.e. user perceives as hang)
- one drawback of robocopy is, it will produce garbled output in some cases. It does not follow the console code page. It follows the system default (i.e. 1252 on your box, instead of 437). We need to solve this, although I don't have a good suggestion offhand. We might want to bundle a little wrapper exe. Here is what I did in the [PulishBuildArtifacts task](https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/PublishBuildArtifactsV1/Invoke-Robocopy.ps1)

damccorm: I changed the retries and incorporated toolrunner.

Concerning the robocopy output, to be clear is this an issue with how the tool displays output to the console or how the files themselves get copied? It looked like the copy itself was working for me (I tested with 'ç').

I'm not really sure to what degree we care about garbled output from the tool itself.  I guess I'm just not seeing the use case where users would have the type of special characters that would be affected in file paths and be upset that they're not displaying correctly - obviously you've run into this before though, is there something I'm missing?

damccorm: After offline discussion changed this to line up with the original comment.

damccorm:
Decided to close based off of offline discussion. Proceeding with the more conservative #531

PR #521, opened 2019-03-27
Mock getVariable, getVariables and setVariable.

hashtagchris:
Adding variable mocking. To avoid breaking existing unit tests, these functions fall through to the task implementation if `answers.variables` is undefined.

This mocking has parity with the task implementation.

- [x]  Variable names normalized.
- [x] Error thrown for multi-line secrets.

### DIFF
--- a/node/mock-answer.ts
+++ b/node/mock-answer.ts
@@ -83,6 +83,12 @@ export class MockAnswers {
             secret = secret || this._variableMap[key].secret;
         }
 
+        if (secret && value && value.match(/\r|\n/)) {
+            if (!this._variableMap.hasOwnProperty('SYSTEM_UNSAFEALLOWMULTILINESECRET') || this._variableMap['SYSTEM_UNSAFEALLOWMULTILINESECRET'].value.toUpperCase() != 'TRUE') {
+                throw new Error('loc_mock_LIB_MultilineSecret');
+            }
+        }
+
         const info: task.VariableInfo = {
             // task.setVariable uses name instead of the normalized key for the VariableInfo.
             name,
--- a/node/mock-answer.ts
+++ b/node/mock-answer.ts
@@ -35,7 +35,6 @@ export type MockedCommand = keyof TaskLibAnswers;
 export class MockAnswers {
     private _answers: TaskLibAnswers | undefined;
     private _variableMap: { [key: string]: task.VariableInfo } | undefined;
-    private _variables: task.VariableInfo[] | undefined;
 
     public initialize(answers: TaskLibAnswers) {
         if (!answers) {
@@ -80,12 +79,12 @@ export class MockAnswers {
 
         // once a secret always a secret
         const key: string = im._getVariableKey(name);
-        if (im._knownVariableMap.hasOwnProperty(key)) {
-            secret = secret || im._knownVariableMap[key].secret;
+        if (this._variableMap.hasOwnProperty(key)) {
+            secret = secret || this._variableMap[key].secret;
         }
 
         const info: task.VariableInfo = {
-            // task.setVariable uses name instead of the normalized key for VariableInfo.
+            // task.setVariable uses name instead of the normalized key for the VariableInfo.
             name,
             value,
             secret
--- a/node/mock-answer.ts
+++ b/node/mock-answer.ts
@@ -1,6 +1,8 @@
 import * as path from 'path';
 import * as fs from 'fs';
+import * as im from'./internal';
 import * as task from './task';
+import { setVariable } from './mock-task';
 
 export interface TaskLibAnswerExecResult {
     code: number,
@@ -21,11 +23,11 @@ export interface TaskLibAnswers {
     osType?: { [key: string]: string },
     rmRF?: { [key: string]: { success: boolean } },
     stats?: { [key: string]: any }, // Can't use `fs.Stats` as most existing uses don't mock all required properties
-    which?: { [key: string]: string },
     variables?: {
-        nonSecrets?: { [key: string]: string },
+        plaintext?: { [key: string]: string },
         secrets?: { [key: string]: string }
-    }
+    },
+    which?: { [key: string]: string }
 }
 
 export type MockedCommand = keyof TaskLibAnswers;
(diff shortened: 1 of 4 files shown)


node/mock-answer.ts
```diff
@@ -22,18 +22,67 @@ export interface TaskLibAnswers {
     rmRF?: { [key: string]: { success: boolean } },
     stats?: { [key: string]: any }, // Can't use `fs.Stats` as most existing uses don't mock all required properties
     which?: { [key: string]: string },
+    variables?: {
+        nonSecrets?: { [key: string]: string },
```
hashtagchris: I considered adding a `fallbackToTask?: boolean` property, so consumers can specify whether they're specifying all defined variables or not. But mock-answer doesn't have access to task, and I didn't see a simple way to honor that.

node/mock-task.ts
```diff
@@ -64,6 +62,32 @@ module.exports.getBoolInput = task.getBoolInput;
 module.exports.getDelimitedInput = task.getDelimitedInput;
 module.exports.filePathSupplied = task.filePathSupplied;
 
+export function getVariable(name: string): string | undefined  {
```
hashtagchris: I'm *not* normalizing the variable name here (upper casing, replacing periods). It's not clear to me if you want all that business logic when you're providing mock answers.

node/mock-answer.ts
```diff
@@ -22,18 +22,67 @@ export interface TaskLibAnswers {
     rmRF?: { [key: string]: { success: boolean } },
     stats?: { [key: string]: any }, // Can't use `fs.Stats` as most existing uses don't mock all required properties
     which?: { [key: string]: string },
+    variables?: {
```
hashtagchris: It seemed odd to require, or even allow, people to mock `getVariable` and `getVariables` separately. So I changed the pattern slightly and take in a collection of variables that will be used by both. `getVariables` needs to know which variables are secret.

node/mock-task.ts
damccorm: I think we should be doing this - we don't want our tests behaving one way (case sensitive, etc...) while the actual agent behaves differently (non-casesensitive, etc...)

node/mock-task.ts
```diff
@@ -64,6 +62,32 @@ module.exports.getBoolInput = task.getBoolInput;
...
+    const variables = mock.getVariableMap();
```
damccorm: Lets check if `name` is null/undefined and [throw if so](https://github.com/Microsoft/azure-pipelines-task-lib/blob/11c9439d4af17e6475d9fe058e6b2e03914d17e6/node/internal.ts#L256).

node/mock-answer.ts
damccorm: Nit: `variables` ahead of `which`, we're sorting by case here.

node/mock-answer.ts
damccorm: Makes sense, I don't think we need that.

node/mock-answer.ts
damccorm: Makes sense to me, I like this pattern more.

node/mock-answer.ts
```diff
@@ -22,18 +22,67 @@ export interface TaskLibAnswers {
...
+                    };
+
+                    this._variableMap[name] = info;
+                    this._variables.push(info);
+                }
+            }
+        }
+    }
+
+    public getVariableMap(): { [key: string]: task.VariableInfo } | undefined {
```
brcrista: TS has special get / set syntax: https://stackoverflow.com/questions/12827266/get-and-set-in-typescript

node/mock-task.ts
```diff
@@ -64,6 +62,32 @@ module.exports.getBoolInput = task.getBoolInput;
...
+
+    if (variables !== undefined) {
+        if (variables[name]) {
```
brcrista: Just checking, does this support mocking a variable to be undefined?

node/mock-answer.ts
```diff
@@ -22,18 +22,67 @@ export interface TaskLibAnswers {
...
+                        secret: false
+                    };
+
+                    this._variableMap[name] = info;
+                    this._variables.push(info);
+                }
+            }
+
+            if (this._answers.variables.secrets) {
+                for (let name in this._answers.variables.secrets) {
```
brcrista: can be `const`

node/mock-answer.ts
```diff
@@ -22,18 +22,67 @@ export interface TaskLibAnswers {
...
         if (!answers) {
             throw new Error('Answers not supplied');
         }
         this._answers = answers;
+
+        if (this._answers.variables) {
+            this._variableMap = {};
+            this._variables = [];
+
+            if (this._answers.variables.nonSecrets) {
```
brcrista: can be `const`

node/mock-answer.ts
```diff
@@ -22,18 +22,67 @@ export interface TaskLibAnswers {
...
+                for (let name in this._answers.variables.nonSecrets) {
```
brcrista: can be `const`

node/mock-answer.ts
brcrista: Naming suggestion: `plaintext`?

brcrista: Also, I think the types here should be

```ts
{ [key: string]: string | undefined }
```

We will want to test task behavior when some variable isn't set.

hashtagchris: `getVariable` will return `undefined` if the variable with name `name` isn't found. There's no fallback logic if variable answers were provided.

Does the real task allow for variables set to undefined? [`setVariable` takes a `string`](https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/node/task.ts#L161) and [VariableInfo](https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/node/task.ts#L204) has a `string` property. Wouldn't `getVariables` [blow up](https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/node/task.ts#L149) if there were undefined values?

brcrista: Yeah, I believe if you `getVariable('SOME_UNDEFINED_ENVIRONMENT_VARIABLE')` the function will return undefined.

hashtagchris: Right. Why do you want to be able to _explicitly_ say:

```json
variables: {
   "plaintext": {
      ...
      "SOME_UNDEFINED_ENVIRONMENT_VARIABLE": undefined
   }
}
```

When the mocked `getVariable` for any variable not listed in the answers will return `undefined` anyway?

node/mock-task.ts
hashtagchris: I added normalization. I mocked `setVariable` as well, because this should probably work when mocking:

```typescript
setVariable("foo", "bar");
const val = getVariable("foo");
```

What haven't done yet is throw on multi-line secrets. If we need to do that too, I'm starting to wonder if we should just have our tests, or mock-answer.ts, call the real `setVariable`. Perhaps we can refactor the variable code so it doesn't access `process.env` when running unit tests.

node/mock-task.ts
hashtagchris: Resolved. `getVariable` is now calling `im._getVariableKey` to throw or normalize.

node/mock-answer.ts
hashtagchris: Good catch, I don't why I thought `const` was forbidden within a `for` loop.

node/mock-task.ts
hashtagchris: You can't explicitly state that variable is undefined. The absence of a variable in the answers collection makes it undefined. I believe ultimately the real `getVariables` only returns defined variables.

Related, if I run `process.env["ZZZ"] = undefined` in node, `process.env["ZZZ"]` actually gets set to string "undefined". `for (const name in process.env) { if (process.env[name] === undefined) { console.log(name) } }` returns no entries.

node/mock-answer.ts
brcrista: You can't do it if you do `i++`

node/mock-answer.ts
brcrista: Yeah, that's what I meant.  I'm under the impression that omitting the mock may or may not return undefined; it will fall through to the actual implementation and depend on whatever's in the process's environment

node/mock-task.ts
hashtagchris: `getVariable`'s behavior is hopefully a little clearer now.

node/mock-answer.ts
hashtagchris: I made a few updates. It should be clearer now that `getVariables` doesn't fall through.

```js
// variables not mocked.
answers: {}
```
and

```json
// variables mocked - getVariable returns undefined for any variable name.
answers: {
  "variables": {}
}
```

node/mock-task.ts
hashtagchris: Added error on multi-line secrets. I may take a look at mocking just the access to `process.env` and `im.vault`.

damccorm:
This is great - my only comment at this point is could we add a couple unit tests in `node/tests/mocktests.ts`?

node/mock-task.ts
hashtagchris: I'm looking into the feasibility of adding real and mock backing stores so we can share logic across task and mock-task.

interfaces.ts (new file)
```ts
export interface SecretStore { // implemented by Vault and some test class.
    storeSecret(name: string, data: string): boolean;
    retrieveSecret(name: string): string | undefined;
}
```

internal.ts
```ts
// export var _vault: vm.Vault;
export var _vault: iface.SecretStore;
```

damccorm:
@hashtagchris any update on this? I think this would be nice to have when we move out of preview

damccorm:
@hashtagchris any updates?

hashtagchris:
@damccorm I'd like to switch to trying out dependency injection, rather than duplicating all the logic (normalizing keys, no multi-line secrets) across two files and having to keep the logic in sync in the future. Unfortunately I may not have any time to try that until May.

stephenmichaelf:
Closing due to idleness and no tests.

PR #526, opened 2019-04-08
Clean environment after running mocks

### LINKED ISSUE #471: Values provied to setInput() appear to be shared across tests
I am noticing that inputs set from the tests appear to carry over from test to test. For example lets say I have an `setInput("backendType", "foo")` in my first test. Then in the second test I am asserting what happens if `backendType` input is not set (its an optional input). When the second test executes the stdout shows that the value of `backendType` is still `foo`. 

This looks to be due to the fact that setInput simply sets a process environment variable. The environment variables seem to carry over from test to test. 

I was able to work around this issue by adding this code to the start of each scenario I am testing
```javascript
//clear any environment vars set by the previous run
Object.keys(process.env)
    .filter(key => key.startsWith("INPUT_"))
    .forEach(key => delete process.env[key]);
```

I was expecting each run to execute in an isolated/clean environment. If this is by design feel free to close this. However, I wanted this documented in case someone else runs across the same issue.

damccorm:
Right now, if you set the input for one test run that carries over to the next one. This disables that behavior

Fixes #471

### DIFF
--- a/node/mock-test.ts
+++ b/node/mock-test.ts
@@ -66,6 +66,14 @@ export class MockTestRunner {
             nodePath = this.getNodePath(nodeVersion);
         }
         let spawn = cp.spawnSync(nodePath, [this._testPath]);
+
+        // Clean environment
+        Object.keys(process.env)
+            .filter(key => (key.substr(0, 'INPUT_'.length) === 'INPUT_' ||
+                            key.substr(0, 'SECRET_'.length) === 'SECRET_' ||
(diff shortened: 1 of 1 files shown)


stephenmichaelf:
Do we have a plan to verify that this does't break all of the task tests? I am not sure if this functionality is expected but it's possible that teams are relying on it.

damccorm:
I don't think we have that guarantee, that's why I was waiting for the major version upgrade

damccorm:
I do think most tests will be fine though.

stephenmichaelf:
Sounds good, let's make sure to update docs/code comments where appropriate. I am fine doing this as part of the major version bump.

PR #529, opened 2019-04-18
Dont perform which twice when check is true

damccorm:
Right now, if `check` is true we get the result of `_which(tool, false)` and then don't use that result if the tool is found. We should just short circuit there if the tool is found, rather than basically performing the `_which` an extra time

### DIFF
--- a/node/internal.ts
+++ b/node/internal.ts
@@ -345,7 +345,10 @@ export function _which(tool: string, check?: boolean): string {
     // recursive when check=true
     if (check) {
         let result: string = _which(tool, false);
-        if (!result) {
+        if (result) {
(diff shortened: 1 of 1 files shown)


stephenmichaelf:
Can we add a test for this?

damccorm:
> Can we add a test for this?

Functionally this shouldn't change behavior, so [existing tests](https://github.com/Microsoft/azure-pipelines-task-lib/blob/1a7e4ddca5277f87301b4880fed1971022c16b7e/node/test/dirtests.ts#L42) should provide the coverage we need. Beyond that, I'm not really sure what we'd be testing.

PR #531, opened 2019-05-01
Fix rmrf and drop version to 2.9.0

damccorm:
Current published version is 2.8.0. Probably not going to major version at this time.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -233,10 +233,11 @@ export function getInput(name: string, required?: boolean): string | undefined {
 /**
  * Gets the value of an input and converts to a bool.  Convenience.
  * If required is true and the value is not set, it will throw.
+ * If required is false and the value is not set, returns false.
  * 
  * @param     name     name of the bool input to get
  * @param     required whether input is required.  optional, defaults to false
- * @returns   string
+ * @returns   boolean
  */
 export function getBoolInput(name: string, required?: boolean): boolean {
     return (getInput(name, required) || '').toUpperCase() == "TRUE";
@@ -421,7 +422,6 @@ export function getEndpointAuthorization(id: string, optional: boolean): Endpoin
         setResult(TaskResult.Failed, loc('LIB_EndpointAuthNotExist', id));
     }
 
-    console.log(id + ' exists ' + (aval !== null));
     debug(id + ' exists ' + (aval !== null));
 
     var auth: EndpointAuthorization | undefined;
--- a/node/task.ts
+++ b/node/task.ts
@@ -1170,6 +1170,7 @@ export function rmRF(inputPath: string): void {
     
     if (getPlatform() == Platform.Windows) {
         if (lstats.isDirectory()) {
+            debug('removing directory');
             childProcess.execSync(`rd /s /q "${inputPath}"`);
         }
         else {
--- a/node/task.ts
+++ b/node/task.ts
@@ -1153,7 +1153,7 @@ function _legacyFindFiles_getMatchingItems(
  */
 export function rmRF(inputPath: string): void {
     debug('rm -rf ' + inputPath);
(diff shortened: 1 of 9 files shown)


stephenmichaelf:
Can we add more details to this? What are we fixing? What's the new implementation? How is it different? Good to have for historical context.

stephenmichaelf:
Are there tests that cover this? Do we need more?

damccorm:
Sure, for context we're trying to solve the issue [here](https://github.com/Microsoft/azure-pipelines-tasks/issues/5472). To summarize, nodes unlinkSync doesn't always actually delete a file, it just decrements its ref-count (which eventually has the effect of deleting it). This means that if the file is open elsewhere (often with antivirus software its being scanned) it doesn't actually get immediately deleted. This tries to address that.

I think existing coverage does a pretty good job with this, not really sure if there's a good way to test that exact behavior. Since we're not using unlinkSync, that edge case should just be avoided entirely anyways.

node/task.ts
```diff
@@ -1150,42 +1151,35 @@ function _legacyFindFiles_getMatchingItems(
...
-        throw new Error(loc('LIB_OperationFailed', 'rmRF', err.message));
     }
-
-    if (lstats.isDirectory()) {
-        debug('removing directory');
-        shell.rm('-rf', path);
-        let errMsg: string = shell.error();
-        if (errMsg) {
-            throw new Error(loc('LIB_OperationFailed', 'rmRF', errMsg));
+    catch (err) {
```
stephenmichaelf: Did we lose the return? Hard to read the diff.

stephenmichaelf:
Looks good, let's have Eric review before merging.

node/task.ts
damccorm: Yeah, we don't actually want that anymore though, we want to try to get rid of the dead symlink regardless

node/task.ts
```diff
@@ -1150,42 +1151,35 @@ function _legacyFindFiles_getMatchingItems(
...
+        if (getPlatform() == Platform.Windows) {
+            if (fs.statSync(inputPath).isDirectory()) {
+                childProcess.execSync(`rd /s /q "${inputPath}"`);
+            }
+            else {
+                childProcess.execSync(`del /f /a "${inputPath}"`);
+            }
+        }
+        else {
+            childProcess.execSync('rm -rf ' + inputPath);
```
damccorm: Use previous implementation for Linux/mac

node/task.ts
```diff
@@ -1150,42 +1151,35 @@ function _legacyFindFiles_getMatchingItems(
...
         }
-
-        return;
     }
-
-    debug('removing file');
     try {
-        fs.unlinkSync(path);
+        // Still try to unlink in case its a dead symlink.
+        fs.unlinkSync(inputPath);
```
damccorm: Don't need this now

node/task.ts
```diff
@@ -1150,42 +1151,77 @@ function _legacyFindFiles_getMatchingItems(
...
-        // if you try to delete a file that doesn't exist, desired result is achieved
-        // other errors are valid
-        if (err.code == 'ENOENT') {
-            return;
+    if (getPlatform() == Platform.Windows) {
+        // Node doesn't provide a delete operation, only an unlink function. This means that if the file is being used by another
+        // program (e.g. antivirus), it won't be deleted. To address this, we shell out the work to rd/del.
+        try {
+            if (fs.statSync(inputPath).isDirectory()) {
+                debug('removing directory');
```
jtpetty: is there a reason to not put the path you are trying to delete in the debug message?

damccorm: Nope, will update

damccorm:
I'm intentionally not updating the version in package.json. We should probably release all the changes going in and need to follow up in the release notes when we do.

PR #536, opened 2019-05-13
Fixed writeFile to be really synchronous

Anumita:
This fix is to make the writeFile really synchronous. fs.writeFileSync is not really synchronous.  This was discovered when customers found that even though the kubectl tool was downloaded it wasn't present in the cached location sometimes. This was a varient issue.

We had made a similar change in azure-pipelines-task repo 
PR: https://github.com/microsoft/azure-pipelines-tasks/pull/6545

This fix ensures that after writing the changes are flushed to disk.
More on this writeFileSync issue:
http://www.daveeddy.com/2013/03/26/synchronous-file-io-in-nodejs/

### DIFF
--- a/node/Strings/resources.resjson/en-US/resources.resjson
+++ b/node/Strings/resources.resjson/en-US/resources.resjson
@@ -30,5 +30,7 @@
   "loc.messages.LIB_OperationFailed": "Failed %s: %s",
   "loc.messages.LIB_UseFirstGlobMatch": "Multiple workspace matches. using first.",
   "loc.messages.LIB_MergeTestResultNotSupported": "Merging test results from multiple files to one test run is not supported on this version of build agent for OSX/Linux, each test result file will be published as a separate test run in VSO/TFS.",
-  "loc.messages.LIB_PlatformNotSupported": "Platform not supported: %s"
+  "loc.messages.LIB_PlatformNotSupported": "Platform not supported: %s",
+  "loc.messages.LIB_FileContentSynced": "Synced the file content to the disk. The content is %s.",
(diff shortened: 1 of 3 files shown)


damccorm:
A couple of thoughts:
1) If I'm understanding correctly, the issue isn't that writeFileSync doesn't end up writing the file synchronously, its that it doesn't necessarily get persisted all the way to the disk - is that correct?
2) If I'm correct in (1), I'm not sure we should be making this change - it would introduce significant overhead, and probably isn't needed in most use cases - should this be the responsibility of the task author?

Anumita:
You are correct about (1). This caching tool though would be used by a lot of tasks.
Customers have been reporting issues that the kubectl task is not stable because of the issue mentioned above.(https://developercommunity.visualstudio.com/content/problem/526764/deploy-to-kubernetes-task-in-azure-devops-on-hoste.html).

Regarding (2), what actions would come under a task author in this case according to you? Should i not us the cachelib provided by task lib and create my own caching function? Just wanted to know what would the correct action would be :)

damccorm:
> what actions would come under a task author in this case according to you?

I think the distinction I'm drawing here is that things that are not needed in all or most use cases shouldn't be part of this library, especially if they introduce significant overhead. So this would be left to the author because it falls in that category (I think). This library should optimize for the common use cases and leave more intensive/unique work to the task authors

stephenmichaelf:
Agree with Danny's assessment here.

It makes sense to write this as code that wraps the lib.

Anumita:
Will do. Thanks for the clarification!

PR #543, opened 2019-06-06
Escape % signs

### LINKED ISSUE #542: Cannot use task.setvariable to define variable containing "%3B".
### Environment
azure-pipelines-task-lib version: 
```
Current agent version: '2.152.1'
```

### Issue Description
When attempting to use `##vso[task.setvariable]` to set a variable to a value containing a literal `%3B`, the variable is instead stored to a value containing `;` as per the unescaping rules at https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/node/taskcommand.ts#L83.

For instance, the following pipeline raises the issue:

```yaml

variables:
  Foo: Bar

pool:
  vmImage: 'ubuntu-latest'

steps:
- powershell: |
    echo ("vso[task.setvariable variable=Abc.Beans;]$(Foo)%" + "3Bdef")
    echo ("##vso[task.setvariable variable=Abc.Lolz;]$(Foo)%" + "3Bdef")
    echo "##vso[task.setvariable variable=Abc.Def;]$(Foo)%253Bdef"
    echo "##vso[task.setvariable variable=Abc.Things;]$(Foo)%25def"
  displayName: 'Run a one-line script'

- script: |
    echo '$(Foo)'
    echo "Something still else changed"
    echo '$(Abc.Beans)'
    echo '$(Abc.Def)'
    echo '$(Abc.Things)'
```

### Expected behaviour
```
echo "Something still else changed"
echo 'Bar%3Bdef'
echo 'Bar%253Bdef'
echo 'Bar%25def'
```

### Actual behaviour
```
echo "Something still else changed"
echo 'Bar;def'
echo 'Bar%253Bdef'
echo 'Bar%25def'
```

damccorm:
This will allow users to pass the literal value `%0A` for example and have it get correctly escaped. See #542 for context

Fixes #542

### DIFF
--- a/node/taskcommand.ts
+++ b/node/taskcommand.ts
@@ -91,9 +91,9 @@ export function commandFromString(commandLine) {
 }
 
 function escapedata(s) : string {
-    return s.replace(/\r/g, '%0D')
-            .replace(/\n/g, '%0A')
-            .replace(/%/g, '%25');
+    return s.replace(/%/g, '%25')
+            .replace(/\r/g, '%0D')
+            .replace(/\n/g, '%0A');
 }
 
 function unescapedata(s) : string {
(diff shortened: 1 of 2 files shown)


stephenmichaelf:
Can we add tests for this? We should go slow with this PR as it's a sensitive area to change.

damccorm:
Yeah, not sure why I opened this PR, it still needs some work. Closing and will reopen with adequate testing.

damccorm:
Ok, I think it should be good to go now.

PR #546, opened 2019-06-30
Add link to a typescript task starter template 

shudv:
**Reason for making this change**

It took me quite a lot of time to setup a working typescript task project. It wasn't obvious how the dependencies are supposed to be packaged within the task. This template has all the build and packaging setup out of the box so that users can get straight to experimenting with their custom logic.

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -11,7 +11,7 @@ Cross platform tasks are written in TypeScript.  It is the preferred way to writ
 
 [![NPM version][npm-lib-image]][npm-lib-url]
 
-Step by Step: [Create Task](https://docs.microsoft.com/en-us/azure/devops/extend/develop/add-build-task?view=vsts)  
+Step by Step: [Create Task](https://docs.microsoft.com/en-us/azure/devops/extend/develop/add-build-task?view=vsts) ([starter template](https://github.com/shudv/azure-pipelines-task-template))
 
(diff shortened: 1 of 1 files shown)


shudv:
I wonder whether or not linking to source from independent authors is allowed/acceptable. If it's not, what should be the correct approach here?

stephenmichaelf:
Thanks for taking the time to create a PR for the repo, unfortunately I don't think we are going to accept it for the reason you stated. It would also require the external repo to immediately keep up with any changes to how tasks are created.

We have a link at the top of the repo to our existing "in the box" tasks as samples of how to get started. There are also docs [here](https://docs.microsoft.com/en-us/azure/devops/extend/develop/add-build-task?view=azure-devops) on how to create a task.

If there's anything specific we can add to the docs please let me know!

shudv:
Thankyou @stephenmichaelf. I understand. I immediately realised the downsides of a change like this as soon as I raised the PR. 

The main point which I think is not clear from the docs is **how should the npm dependencies be packaged into the task**. Questions like - At what level should they be placed or should they be bundled together instead etc. At least for someone who is not very experienced working with TypeScript/Node could benefit from a quick-start template which has build and packaging setup by default.

I think a good middle ground would be to beef up the step by step instructions and make the dependency handling point clearer. I could do it but I don't think I would do a good job of it as I haven't updated the official docs ever before. I leave it to your judgement.

PR #550, opened 2019-07-05
Add Optional Done Parameter to Task Complete Message

thboop:
Added the ability to specify done on a powershell job via the sdk, similar to how the node sdk works

### DIFF
--- a/powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
+++ b/powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
@@ -202,7 +202,7 @@ function Write-SetResult {
         [string]$Result,
         [string]$Message,
         [Parameter()]
-        [bool]$Done,
+        [switch]$Done,
         [Parameter(ParameterSetName = 'AsOutput')]
         [switch]$AsOutput,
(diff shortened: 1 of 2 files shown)


powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
```diff
@@ -201,14 +201,17 @@ function Write-SetResult {
         [Parameter(Mandatory = $true)]
         [string]$Result,
         [string]$Message,
+        [Parameter()]
+        [bool]$Done,
```
ericsciple: use `[switch]` (true to the powershell way) instead of `[bool]`

thboop: I'll implement this change!

ericsciple: thanks. otherwise this lgtm.

Make sure the E2E scenario works before you merge.

Then after you merge this, sync w/ danny to release the new lib version, and consume the new version in the command line task and powershell task.

thboop:
Closing this as we may need to look into another approach

PR #563, opened 2019-08-16
Update Linux icon

mrbcmorris:
Hey folks, I navigated here from a azure devops build warning and noticed the icon for linux was an ubuntu icon.  Thought it might be a bit more appropriate to give ambassador tux some love in this situation.

Let me know what you think!
Chris


damccorm:
I actually think I like having the Ubuntu icon since the flavor of Linux we're using is Ubuntu - so it adds more info rather than less.

PR #570, opened 2019-09-09
Updated task variables to support dots

### LINKED ISSUE #303: Having a dot in a task variable name is not supported

damccorm:
Same as #414, my fork got deleted though and this is easier.

Fixes #303

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -217,7 +217,7 @@ export interface VariableInfo {
  * @returns   string
  */
 export function getInput(name: string, required?: boolean): string | undefined {
-    var inval = im._vault.retrieveSecret('INPUT_' + name.replace(' ', '_').toUpperCase());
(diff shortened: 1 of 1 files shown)


damccorm:
I'm going to just merge since I got an approval in #414

PR #575, opened 2019-09-23
Added .npmigore file to not include tests in npm package

jtpetty:
When I ran npm publish, the created package contained test directories which has a task.json which breaks tests in azure-pipelines-task. This excludes those directories from getting in the package.


jtpetty:
This is not needed as we should be publishing from _build directory

PR #582, opened 2019-10-25
docs: Include required file placement in docs

deadlydog:
Previously the task execution target file did not have to be in the task root directory; it could be in a subdirectory and be referenced by the `task.json` file. With the new `PowerShell3` task execution handler, this is no longer allowed as it results in a "file not found" error at runtime.

### DIFF
--- a/powershell/Docs/Consuming.md
+++ b/powershell/Docs/Consuming.md
@@ -42,3 +42,5 @@ MyTask
             [...]
             VstsTaskSdk.psd1
 ```
+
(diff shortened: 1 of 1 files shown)


jtpetty:
@damccorm - can you take a look at this one?

damccorm:
I don't think this should be true - you should be able to specify the location with the `target` command - is this not working for you? If so, please file an issue and we can address it that way

deadlydog:
Ok, I'll file a bug for this then. You can see the directory structure and `target` command being used [in this StackOverflow question I posted regarding this](https://stackoverflow.com/questions/58527792/azure-devops-custom-build-task-extension-cannot-find-powershell-vststasksdk-psd1). Thanks.

PR #588, opened 2019-11-14
Add missing assertion in single pattern findMatch tests

kfcampbell:
Hi! I was investigating a customer issue over at App Center that required some debugging with findMatch. While the error ended up not being related to this task, I did notice a missing test assertion while I was looking into it.

I think this is just an oversight, so here's a quick PR to add it!


damccorm:
Thanks for catching this!

kfcampbell:
Yeah of course, happy to help!

PR #599, opened 2020-01-08
Update scripts.

stephenmichaelf:
Trying to get consistency in our CI scripts. I started with the tool lib and am copying that setup here.

Let me know if you have thoughts on this/want to change anything.

Planning to do this for all OSS libs.

### DIFF
--- a/azure-pipelines-steps.yml
+++ b/azure-pipelines-steps.yml
@@ -1,0 +1,17 @@
+
+steps:
+
+- script: npm install
+  displayName: 'npm install'
+
+- script: npm run build
+  displayName: 'npm run build'
+
+# test with node 6
+- task: NodeTool@0
+  displayName: Use Node 6.10.3
+  inputs:
+    versionSpec: "6.10.3"  
+
+- script: npm test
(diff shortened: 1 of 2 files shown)


azure-pipelines-steps.yml
```diff
@@ -0,0 +1,17 @@
...
+steps:
+
+- script: npm install
+  displayName: 'npm install'
+
+- script: npm run build
+  displayName: 'npm run build'
+
+# test with node 6
+- task: NodeTool@0
```
damccorm: I think if we do this we should matrix across different versions of node - probably 6 and 12, maybe 5 too (we currently do 5). Note that this need isn't consistent across all our OSS repos, e.g. in typed-rest-client we currently test as low as 4 I believe

damccorm: We might also want to matrix across more hosted platforms? E.g. windows-2019

damccorm:
Also, this failed CI with a parse error

stephenmichaelf:
Will reopen when I can look at this.

PR #613, opened 2020-01-23
Annotate outStream and errStream as optional

AArnott:
In fact the handling of these properties already allows for them to be undefined/null. But the annotations didn't indicate this.
Since the entire options object was optional, having a couple properties on it *not* optional makes it hard to start supplying *any* options since the user must figure out what values to supply for these two properties to get equivalent behavior, in order to set some *other* option on the object.

### DIFF
--- a/node/toolrunner.ts
+++ b/node/toolrunner.ts
@@ -32,9 +32,11 @@ export interface IExecSyncOptions {
     /** optional.  defaults to false */
     silent?: boolean;
 
-    outStream: stream.Writable;
+    /** Optional. Default is process.stdout. */
+    outStream?: stream.Writable;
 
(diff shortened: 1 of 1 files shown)


AArnott:
If someone with push permissions can tell me they'd take this  PR, I'll work out the build failures.

damccorm:
Hey, sorry for the slow response - I'll merge this if you fix up the build failures.

AArnott:
Thanks, @damccorm. I fixed it, but the windows agent failed due to what appears to be an unrelated network issue. Can you confirm and resolve?

damccorm:
Yeah, that's clearly unrelated - I'll merge here and follow up on that separately. Thanks for doing this!

Issue #617, opened 2020-02-03
Support multiple pipe redirection


lstocchi:
### Environment
azure-pipelines-task-lib version: 2.9.3

### Issue Description
I'm trying to pipe stdout of my first tool in a sequence of N tools but it only works on the first pipe operator.
Let's suppose i have `cmd1 | cmd2 | cmd3` i would like to pass the cmd1 stdout to cmd2 and then cmd2 stdout to cmd3.

### Expected behaviour
Let's suppose i have `cmd1 | cmd2 | cmd3` and i create a toolrunner `cmd1.pipeExecOutputToTool (cmd2.pipeExecOutputToTool(cmd3))` i would like it to pass the cmd1 stdout to cmd2 and then cmd2 stdout to cmd3.

### Actual behaviour
Let's suppose i have `cmd1 | cmd2 | cmd3` it only executes cmd1 and cmd2.

PR #618, opened 2020-02-07
Handle multiple pipes

lstocchi:
This is a POC for #617 

This PR allows the lib to handle multiple pipes. I don't know if there is some reason why it was created to support only one pipe but i created this example so we can discuss about it. 

Until now if i run a cmd like `oc describe pod/nodejs-ex-3-xtnqn | grep node | grep kubernetes` it resulted in:
image

With this PR the output is:
image

This is a very simple example but you can see the limitation of just allowing one single pipe.

I edited the code as less as i could waiting to collaborate with you because i had many doubts. I.e if we allow to handle multiple pipes and a filename is passed, which data we should write to the file? The final result after all pipes run or just the first toolrunner's result?

Thank you

### DIFF
--- a/node/test/toolrunnertests.ts
+++ b/node/test/toolrunnertests.ts
@@ -631,7 +631,7 @@ describe('Toolrunner Tests', function () {
                 .arg('line 1')
                 .arg('line 2')
                 .arg('line 3');
-            outputExe.pipeExecOutputToTool(matchExe);
+            outputExe.pipeExecOutputToTool([matchExe]);
 
             var output = '';
             outputExe.on('stdout', (data) => {
@@ -654,7 +654,7 @@ describe('Toolrunner Tests', function () {
 
             var ps = tl.tool(tl.which('ps', true));
             ps.arg('ax');
-            ps.pipeExecOutputToTool(grep);
+            ps.pipeExecOutputToTool([grep]);
 
             var output = '';
             ps.on('stdout', (data) => {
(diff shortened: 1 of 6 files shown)


node/test/toolrunnertests.ts
```diff
@@ -1841,7 +1841,7 @@ describe('Toolrunner Tests', function () {
             cmdRunner.on('stdout', (data) => {
                 output += data.toString();
             });
-            cmdRunner.pipeExecOutputToTool(exeRunner);
+            cmdRunner.pipeExecOutputToTool([exeRunner]);
             cmdRunner.exec(options)
```
damccorm, line 1845: Could you add a new test for multi-piping?

node/toolrunner.ts
```diff
@@ -183,8 +183,10 @@ export class ToolRunner extends events.EventEmitter {
         }
 
         // append second tool
```
damccorm, line 185: Nit: update comment

node/toolrunner.ts
```diff
@@ -764,12 +771,12 @@ export class ToolRunner extends events.EventEmitter {
...
     /**
      * Pipe output of exec() to another tool
-     * @param tool
+     * @param tools
      * @param file  optional filename to additionally stream the output to.
      * @returns {ToolRunner}
      */
-    public pipeExecOutputToTool(tool: ToolRunner, file?: string): ToolRunner {
-        this.pipeOutputToTool = tool;
+    public pipeExecOutputToTool(tools: ToolRunner[], file?: string): ToolRunner {
```
damccorm, line 778: I think we shouldn't break back compat here. We should still allow you to call `pipeExecOutputToTool(tool)` - that can turn around and immediately call `pipeExecOutputToTool([tool])

lstocchi:
I completely forgot to look into this. Sorry. Thank for the feedback @damccorm. I'll work on this in next days and push a proper fix.

PR #632, opened 2020-05-05
Update Docs to point to a simple and more clear example 

Angr1st:
I personally found the linked build task pretty hard to follow because some of the folder structure was replaced during build time/by some script. I updated my repo to contain a very simple and straightforward example of a build task using the powershell3 task executer. This could be helpful for people with a similar goal than I had. Which was to just package another executable on windows to be run as a build task.

### DIFF
--- a/powershell/Docs/README.md
+++ b/powershell/Docs/README.md
@@ -6,7 +6,7 @@ The VSTS Task SDK for PowerShell is designed to work with the agent's new PowerS
 ## Reference Examples
 
 The [CustomBuildTask](https://github.com/Angr1st/CustomBuildTask) is a really simple example that you can use as a template.
-If you want to see more complicated but better made examples you can have a look at the [MSBuild Task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/MSBuildV1/MSBuild.ps1) and [VSBuild Task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/VSBuildV1/VSBuild.ps1).
+If you want to see more complicated but more well made examples you can have a look at the [MSBuild Task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/MSBuildV1/MSBuild.ps1) and [VSBuild Task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/VSBuildV1/VSBuild.ps1).
 
 ## Documentation
 
--- a/powershell/Docs/README.md
+++ b/powershell/Docs/README.md
@@ -5,7 +5,8 @@ The VSTS Task SDK for PowerShell is designed to work with the agent's new PowerS
 
 ## Reference Examples
 
-The [MSBuild Task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/MSBuildV1/MSBuild.ps1) and [VSBuild Task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/VSBuildV1/VSBuild.ps1) are good examples.
+The [CustomBuildTask](https://github.com/Angr1st/CustomBuildTask) is a really simple example that you can use as a template.
+If you want to see more complicated but better made example you can have a look at the [MSBuild Task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/MSBuildV1/MSBuild.ps1) and [VSBuild Task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/VSBuildV1/VSBuild.ps1).
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


Angr1st:
[Build Erros](https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=12068996&view=logs&jobId=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=be5e7cb1-b8ca-569b-a02f-a5442c8a3f8a) look like something ran into a timeout. DISCLAIMER: I'm not a js expert. I could be totally off but I haven't touched anything but the one README.md so I'm pretty sure your tests are potentially a little flaky.

Angr1st:
You are also using the [old build agent image](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/azure-pipelines.yml#L12) for windows. You should update to `windows-2019`.

Angr1st:
The useNode BuildTask is not a BuildTask that uses the VstsTaskSdk so it is
not a valid example for a simple build task using it. It is totally
understandable that you don't want to link to external sources. Perhaps you
could just fork my repo?

Danny McCormick <notifications@github.com> schrieb am Do., 7. Mai 2020,
17:30:

damccorm:
> The useNode BuildTask is not a BuildTask that uses the VstsTaskSdk

Whoops, lost context on what part of the doc we were in - you're right. Maybe we can use the [cmdLine task](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/CmdLineV2/cmdline.ps1)

> Perhaps you could just fork my repo?

I think I'd rather avoid adding another repo to accomplish this since we do have examples that exercise this. It might be helpful to understand what you found confusing about these examples as well.

Angr1st:
Your task all are always cluttered with .ts and .js and package.json files
that are not needed for a windows only build task. Also you use make.js to
copy data during build time of your task. That makes it hard to simply copy
them and be done with it. For me it was just intransparent how the build
task looks once all scrips have run and it is packed as .vsix.
Mine on the other hand does no build time creating/changing the folder at
all. That means you can clone my repo and follow the readme.md to have a
build task ready. Where as with yours I would need to find out how to build
them first. That takes more time.


Danny McCormick <notifications@github.com> schrieb am Do., 7. Mai 2020,
18:03:

damccorm:
That's fair, however, these docs are specifically mean to demonstrate consuming the sdk, not setting up the larger task. That's why we link to the ps1 files, not the task folder. We have other docs that specifically talk about the setup and address that.

I think if we're trying to solve that problem, I'd prefer linking to [the tutorial](https://docs.microsoft.com/en-us/azure/devops/extend/develop/add-build-task?view=azure-devops#preparation-and-required-setup-for-this-tutorial) over adding another repo to maintain

Angr1st:
You are right about that. I guess I thought you were just presenting example of how a build task should look like in the end.

Angr1st:
Instead of linking to another repo I just created an example folder structure inside of the Docs folder. The content of the task.json and task.ps1 should be changed to some microsofty default like Fabrikam or something or what do you think?

damccorm:
I'm ok with this approach, but my main issue here is that we don't ever recommend checking powershell modules or binaries into a repo. The pattern we use (and what we'd recommend) is having a pipeline that builds the task, adding in executables as necessary (and that ultimately can publish the task). So I'd be worried about people just copying the repo.

The reasoning behind that is that in any project that has multiple users, checking in binaries introduces a security risk - its extremely challenging to verify the contents of binaries are actually what they should be.

In general, I'd be more in favor of simply making the docs about structure clearer over trying to provide an in-repo example (or even pointing to a stripped down example)

jessehouwing:
In that case, why isn't there a pre-made script that stores the pipeline
sdk in the right way. I know of no custom task writer who doesn't check in
the binaries apart from Microsoft.

For GitHub Actions we also need to commit the SDK directly to the task repo
for it to load..

damccorm:
> In that case, why isn't there a pre-made script that stores the pipeline
sdk in the right way.

It doesn't exist because nobody's taken the time to do that :) Definitely not opposed to including something like that in the docs. I don't currently have the bandwidth to do that, but I'm happy to review a pr and/or feel free to open up an issue.

> For GitHub Actions we also need to commit the SDK directly to the task repo
for it to load

Can't totally speak to that since its a different product (admittedly with some shared stuff), but I believe this is actually not true anymore. They updated their publishing model to use GitHub releases (in large part because of this problem). For example, no node_modules are checked into this repo - https://github.com/actions/stale

Angr1st:
Is there a module manager for powershell? Could you use nuget/paket to ensure reproducable builds (e.g. lock the sdk module version) using a package.lock file? That would be the best way to get the modules at build time I think.

Issue #630, opened 2020-04-25
Mock Test Runner fails to download a new NodeJS distribution


ElderMael:
Please check our current Issues to see if someone already reported this https://github.com/Microsoft/azure-pipelines-task-lib/issues

### Environment
azure-pipelines-task-lib version: 2.9.3

### Issue Description
While executing unit tests, when calling `MockTestRunner.run()` a binary distribution of NodeJS is being downloaded. While running on a `ubuntu-latest` agent, the download fails without any information error.

### Expected behaviour
The NodeJS binary is downloaded successfully. The test runs.

### Actual behaviour
An error is thrown during the test with the following message:
```
      Downloading file: https://nodejs.org/dist/v10.15.1/node-v10.15.1-linux-x64.tar.gz
```

### Steps to reproduce
1. Clone https://github.com/codexarcana/azure-tasks
1. Create an azure pipeline with the following code:
    ```
        npm install
        npm run test
    ```

### Logs
```bash
FAIL tests/integration.test.ts
  ● Console

    console.log
      Downloading file: https://nodejs.org/dist/v10.15.1/node-v10.15.1-linux-x64.tar.gz

      at MockTestRunner.downloadFile (node_modules/azure-pipelines-task-lib/mock-test.js:228:17)

  ● should succeed with simple inputs



       6 | 
       7 |     const testPath = path.join(__dirname, 'success.integration.js');
    >  8 |     const testMockRunner = new MockTestRunner(testPath);
         |                            ^
       9 | 
      10 |     testMockRunner.run(10);
      11 | 

      at doRequest (node_modules/sync-request/index.js:27:11)
      at MockTestRunner.downloadFile (node_modules/azure-pipelines-task-lib/mock-test.js:230:22)
      at MockTestRunner.downloadTarGz (node_modules/azure-pipelines-task-lib/mock-test.js:242:14)
      at MockTestRunner.downloadNode (node_modules/azure-pipelines-task-lib/mock-test.js:206:22)
      at MockTestRunner.getNodePath (node_modules/azure-pipelines-task-lib/mock-test.js:147:25)
      at new MockTestRunner (node_modules/azure-pipelines-task-lib/mock-test.js:28:30)
      at tests/integration.test.ts:8:28
      at step (tests/integration.test.ts:33:23)
      at Object.next (tests/integration.test.ts:14:53)
      at tests/integration.test.ts:8:71

```

alex-peck:
```
[...]

PR #638, opened 2020-05-18
Permit getDelimitedInput to split on either a string or a Regexp

MorganRodgers:
Changing the type for `delimiter` to `string | Regexp` in `getDelimitedInput` adds support for multiple delimiters. Supporting multiple delimiters would allow https://github.com/microsoft/azure-pipelines-tasks to which is needed for https://github.com/microsoft/azure-pipelines-tasks/issues/12927 by passing a regular expression to `getDelimitedInput` instead of naively using `split` which breaks when there are no Git tags.

### DIFF
--- a/node/test/inputtests.ts
+++ b/node/test/inputtests.ts
@@ -620,6 +620,18 @@ describe('Input Tests', function () {
 
         done();
     })
+    it('gets delimited input with a Regexp', function (done) {
+        this.timeout(1000);
+
+        var inputValue = 'a,b\nc';
+        process.env['INPUT_DELIM'] = inputValue;
+        im._loadData();
+
+        var outVal = tl.getDelimitedInput(/[,\n]/, ' ', /*required*/false);
+        assert.equal(outVal.length, 3, 'should return array with 3 elements');
+
+        done();
+    })
 
     // getPathInput tests
     it('gets path input value', function (done) {
--- a/node/test/inputtests.ts
+++ b/node/test/inputtests.ts
@@ -627,7 +627,7 @@ describe('Input Tests', function () {
         process.env['INPUT_DELIM'] = inputValue;
         im._loadData();
 
-        var outVal = tl.getDelimitedInput(/[,\n]/, ' ', /*required*/false);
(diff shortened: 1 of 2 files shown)


node/test/inputtests.ts
```diff
@@ -620,6 +620,18 @@ describe('Input Tests', function () {
...
         done();
     })
+    it('gets delimited input with a Regexp', function (done) {
+        this.timeout(1000);
+
+        var inputValue = 'a,b\nc';
+        process.env['INPUT_DELIM'] = inputValue;
+        im._loadData();
+
+        var outVal = tl.getDelimitedInput(/[,\n]/, ' ', /*required*/false);
```
damccorm, line 630: I think this should be `tl.getDelimitedInput('delim', /[,\n]/, /*required*/false);` - that should fix the failing test I believe

damccorm:
Thanks for the contribution! This change generally looks good to me, if you fix up the failing test I'll merge

node/test/inputtests.ts
damccorm: Looks like you still have an extra input (the `' '` should be removed)

MorganRodgers:
Sorry for the flailing. Whenever I run `npm test` I get an error even on master:

```
rodgersm$ npm test

> azure-pipelines-task-lib@2.9.5 test /Users/rodgersm/Projects/azure-pipelines-task-lib/node
> node make.js test

> tsc --outDir /Users/rodgersm/Projects/azure-pipelines-task-lib/node/_build
Downloading file: https://nodejs.org/dist/v6.17.1/node-v6.17.1-darwin-x64.tar.gz
/Users/rodgersm/Projects/azure-pipelines-task-lib/node/node_modules/sync-request/index.js:27
    throw new Error(res.stderr.toString());
    ^

Error
    at doRequest (/Users/rodgersm/Projects/azure-pipelines-task-lib/node/node_modules/sync-request/index.js:27:11)
```

Is it possible to test locally?

damccorm:
It should be, looks like maybe you're running into https://github.com/microsoft/azure-pipelines-task-lib/issues/630.

Regardless, looks good now (there are some failing unrelated powershell tests that need to be addressed separately).

PR #637, opened 2020-05-12
Added shell property to IExecOptions

egor-bryzgalov:
Added shell property to IExecOptions to allow this property to be specified in pipelines tasks code. This option allows handling variables in command strings.

Related to the [issue.](https://github.com/microsoft/azure-pipelines-tasks/issues/12848)

These changes are required for the [PR in azure-pipelines-tasks repository](https://github.com/microsoft/azure-pipelines-tasks/pull/12880).

PR #643, opened 2020-06-10
[ToolRunner] Enhanced 'shell' execution option

egor-bryzgalov:
Enhancement for #637 

Required to fix issue [#12848](https://github.com/microsoft/azure-pipelines-tasks/issues/12848).
Required for PR [#13099](https://github.com/microsoft/azure-pipelines-tasks/pull/13099)

Applied enhancement for `shell` option to wrap a tool path and arguments with double quotes to ensure that command is handled correctly if it is executed inside shell. The path and arguments are wrapped only if `shell` property of `IExecOptions` object set to true. 
Added tests.

### DIFF
--- a/node/test/toolrunnertests.ts
+++ b/node/test/toolrunnertests.ts
@@ -2224,7 +2224,7 @@ describe('Toolrunner Tests', function () {
                 });
                 statRunner.exec(_testExecOptions).then(function (code) {
                     assert.equal(code, 0, 'return code of stat should be 0');
-                    assert.equal(output, '-TEST1=test value;test -TEST2=/one/two/three -TEST3=out:$TEST');
+                    assert.equal(output, '-TEST1=test value;test -TEST2=/one/two/three -TEST3=out:$TEST\n');
                     done();
                 })
                 .fail(function (err) {
--- a/node/test/toolrunnertests.ts
+++ b/node/test/toolrunnertests.ts
@@ -2097,9 +2097,9 @@ describe('Toolrunner Tests', function () {
                 assert.equal(ret.stdout.trim(), 'args[0]: \'test value\'', 'Command should return \"args[0]: \'test value\'\"');
             }
             else {
-                var ret = tl.execSync('stat', '--format "%n" $TESTPATH', _testExecOptions);
+                var ret = tl.execSync('stat', '$TESTPATH', _testExecOptions);
                 assert.equal(ret.code, 0, 'return code of stat should be 0');
-                assert.equal(ret.stdout.trim(), tempPath, `Command should return \'${tempPath}\'`);
+                assert(ret.stdout.includes(tempPath), `Result should include \'${tempPath}\'`);
             }
     
             assert(ret.stdout && ret.stdout.length > 0, 'should have emitted stdout');
@@ -2127,13 +2127,13 @@ describe('Toolrunner Tests', function () {
             }
             else {
                 let statRunner = tl.tool('stat');
-                statRunner.line('--format "%n" $TESTPATH');
+                statRunner.line('$TESTPATH');
                 statRunner.on('stdout', (data) => {
                     output = data.toString();
                 });
                 statRunner.exec(_testExecOptions).then(function (code) {
                     assert.equal(code, 0, 'return code of stat should be 0');
-                    assert.equal(output.trim(), tempPath, `Command should return \'${tempPath}\'`);
+                    assert(output.includes(tempPath), `Result should include \'${tempPath}\'`);
                     done();
                 })
                 .fail(function (err) {
--- a/node/test/toolrunnertests.ts
+++ b/node/test/toolrunnertests.ts
@@ -153,91 +153,6 @@ describe('Toolrunner Tests', function () {
                 });
         }
     })
-    it('Exec sync inside shell', function (done) {
-        this.timeout(10000);
-
-        let tempPath: string = testutil.getTestTemp();
-        var _testExecOptions = <trm.IExecOptions>{
-            cwd: __dirname,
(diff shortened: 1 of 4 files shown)


node/test/toolrunnertests.ts
```diff
@@ -2062,6 +2062,179 @@ describe('Toolrunner Tests', function () {
...
+                .fail(function (err) {
+                    done(err);
+                });
+            }
+        });
+        it('Exec pipe output to another tool inside shell, succeeds if both tools succeed', function (done) {
+            this.timeout(30000);
+
+            if (os.platform() === 'win32') {
+                var matchExe = tl.tool(compileMatchExe())
```
alexander-smolyakov, line 2149: Could you please use `let` or `const` instead of `var`?

node/toolrunner.ts
```diff
@@ -283,17 +319,90 @@ export class ToolRunner extends events.EventEmitter {
...
+     * @param arg The cmd command arg.
+     * @param additionalChars Additional chars which should be also checked.
+     */
+    private _needQuotesForCmd(arg: string, additionalChars?: string[] | string): boolean {
+        let specialChars: string[] = this.cmdSpecialChars;
+        if (additionalChars) {
+            specialChars = this.cmdSpecialChars.concat(additionalChars);
+        }
+        for (let char of arg) {
+            if (specialChars.some(x => x == char)) {
```
anton-erofeev, line 399:
```suggestion
            return specialChars.some(x => x == char);
```

alexander-smolyakov, line 399: Maybe better use strict equality in that case?

node/toolrunner.ts
```diff
@@ -216,13 +215,50 @@ export class ToolRunner extends events.EventEmitter {
...
+        return arg;
+    }
+
+    /**
+     * Unwraps an arg string wrapped with specified char
+     * @param arg Arg wrapped with specified char
+     * @param wrapChar A char to be removed
+     */
+    private _unwrapArg(arg: string, wrapChar: string): string {
+        if (this._isWrapped(arg, '"')) {
```
damccorm, line 237: Should this be `isWrapped(arg, wrapChar)`?

node/toolrunner.ts
```diff
@@ -216,13 +215,50 @@ export class ToolRunner extends events.EventEmitter {
...
+            const pattern = new RegExp(`(^\\\\?${wrapChar})|(\\\\?${wrapChar}$)`, 'g');
+            return arg.replace(pattern, '');
```
damccorm, line 239: Should this be `arg.trim().replace(pattern, '');`

node/toolrunner.ts
```diff
@@ -283,17 +319,90 @@ export class ToolRunner extends events.EventEmitter {
...
+     * @param arg String to escape char in
+     * @param chars Char should be escaped
+     * @param escapePrefix Char which should be used to escape. "\" is used if not specified.
+     */
+    private _escapeChar(arg: string, charToEscape: string, escapePrefix?: string): string {
+        const escChar: string = escapePrefix || "\\";
+        let prevChar: string = '';
+        let output: string = '';
+        for (const char of arg) {
+            if (char === charToEscape && prevChar !== escChar) {
```
damccorm, line 359: We might need to be a little smarter than this. For example, if you call this function with the input `_escapeChar('\\\\abc', 'a')`, it returns `'\\\\abc'` , I think you want it to return `'\\\\\\abc'`. We could get around this by walking the string left to right and keeping track of whether we're escaped at the moment or not.

node/toolrunner.ts
```diff
@@ -283,17 +319,90 @@ export class ToolRunner extends events.EventEmitter {
...
+        return output;
+    }
+
+    /**
+     * Escape multiple string characters.
+     * @param arg String to escape chars in
+     * @param chars Chars should be escaped
+     * @param escapePrefix Char which should be used to escape. "\" is used if not specified.
+     */
+    private _escapeChars(arg: string, chars: string[], escapePrefix?: string): string {
```
damccorm, line 375: I don't think we use this function anywhere

node/toolrunner.ts
```diff
@@ -283,17 +319,90 @@ export class ToolRunner extends events.EventEmitter {
...
         }
 
         return this.args;
     }
 
+    /**
+     * Escape specified character.
+     * @param arg String to escape char in
+     * @param chars Char should be escaped
+     * @param escapePrefix Char which should be used to escape. "\" is used if not specified.
```
damccorm, line 352: This feels a little overparameterized - do we anticipate this ever using an escape prefix other than `\`? Parameterizing `charToEscape` seems more reasonable because I could see it being `"` or `'` very easily for example

node/toolrunner.ts
```diff
@@ -152,38 +153,36 @@ export class ToolRunner extends events.EventEmitter {
...
             // Windows + verbatim
             else if (options.windowsVerbatimArguments) {
-                cmd += `"${toolPath}"`;
-                args.forEach((a: string): void => {
-                    cmd += ` ${a}`;
-                });
+                commandParts.push(`"${toolPath}"`);
+                commandParts = commandParts.concat(args);
+            }
+            else if (options.shell) {
```
damccorm, line 168: Shell should be added to the IExecOptions interface, right?

egor-bryzgalov: @damccorm Yes. It was added in #637.

damccorm: Oh right, sorry!

egor-bryzgalov:
@damccorm Thank you very much for the review! I've applied changes in accordance with your review points so could you please take a look?

PR #650, opened 2020-07-07
Allowing env variables to be set without forcing VSTS_TASKVARIABLE_ prefix

jeffw16:
I'm currently developing an internal Microsoft ADO task that takes in an output variable from another task. It would be very helpful to be able to insert an output variable within the testing framework as well.


jeffw16:
Any update on whether this could be integrated into the library please?

PR #652, opened 2020-07-22
Dont trim whitespace from inputs

damccorm:
As of now, this doesn't matter because the agent does this anyways. We're adding an option to not do that from the agent and should not then enforce it here.

Agent PR - https://github.com/microsoft/azure-pipelines-agent/pull/3051
Ticket motivating this - https://developercommunity.visualstudio.com/content/problem/1000335/azure-pipelines-trims-leading-and-trailing-rn-from.html

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -1,14 +1,13 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "2.9.6",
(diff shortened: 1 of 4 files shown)


damccorm:
CI failures look unrelated - they're related to the powershell lib

PR #660, opened 2020-09-10
Fix stdout/stderr types

alex-peck:
These types are what stdout and stderr are, not stream.Writeable. New TS versions are complaining about type mismatches


damccorm (approved):
LGTM

PR #665, opened 2020-09-17
Fully support Node 14 and TS 4

alex-peck:
Creates a new semver major version (3.0.0)

* Node 14
* TS 4.0.2
* Remove typings -> replace with `@types`

This PR targets `releases/3.x`. 2.x will remain on `master` until Node 14 has been adopted by the majority of our first party tasks and has been proven stable.

Important changes should be applied to both `master` and `releases/3.x`


alex-peck:
See my comment here: https://github.com/microsoft/azure-pipelines-task-lib/pull/662#issuecomment-694474604

node/make.js
```diff
@@ -48,6 +46,7 @@ target.test = function() {
     cp('-Rf', rp('test/scripts'), testPath);
     cp('-Rf', rp('test/fakeTasks'), testPath);
     process.env['TASKLIB_INPROC_UNITS'] = '1'; // export task-lib internals for internal unit testing
+    set('+e');
```
stephenmichaelf, line 49: What's this for? Can we add a comment?

node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "2.11.0",
+  "version": "3.0.0",
```
stephenmichaelf, line 3: What about fixes for tasks that are still on Node 6?

node/test/dirtests.ts
```diff
@@ -1826,7 +1826,7 @@ describe('Dir Operation Tests', function () {
 
         var worked: boolean = false;
         try {
-            tl.mv(sourceFile, destFile);
+            tl.mv(sourceFile, destFile, "-n");
```
stephenmichaelf, line 1829: What's this for?

node/package.json
alex-peck: Node 6 will remain on the 2.x version, which is why this is targeting releases/3.x instead of master. Both versions will be maintained until we migrate in the box tasks to node 14.

node/make.js
alex-peck: Basically it restores the old behavior of shelljs before the upgrade. It avoids throwing an exception when tests fail and instead just reports that it failed.

node/test/dirtests.ts
alex-peck: Shelljs changed the default for mv from "-n" (don't override files) to "-f" (override files). This restores the idea behind the test. 

I checked in the tasks repo and no tasks are depending on this behavior.

res/UseNode5.ps1
```diff
@@ -1,18 +0,0 @@
-$ErrorActionPreference='Stop'
```
stephenmichaelf, line 1: Are we sure we will never need this?

alex-peck: It was only used in CI and we don't need to run against Node 5 anymore

node/package.json
damccorm: If we're breaking compat we should probably include a writeup somewhere in the repo that describes the upgrade steps needed

damccorm: Also, should probably mention that currently any changes should go into both master and this branch

node/task.ts
```diff
@@ -539,28 +539,22 @@ export interface FsStats extends fs.Stats {
...
-    encoding?: string;
-    mode?: number;
-    flag?: string;
-}
-
-export function writeFile(file: string, data: string | Buffer, options?: string | FsOptions) {
-    if(typeof(options) === 'string'){
-        fs.writeFileSync(file, data, {encoding: options});
+export function writeFile(file: string, data: string | Buffer, options?: BufferEncoding | fs.WriteFileOptions) {
+    if (typeof(options) === 'string'){
```
damccorm, line 556: With the updated input types, is it ever possible for this to return true?

node/package.json
damccorm: Oh, also we might want to release it as a preview version for now (3.0.0-preview) until we fully move away from our 2.x versions and are ready for it to just all be node 14 from here out

node/task.ts
alex-peck: `BufferEncoding` is a type that is a list of possible strings. typeof() should still resolve to `string`

damccorm (approved):

PR #678, opened 2020-10-22
Change type of the resultFiles parameter of publish method in TestPublisher class

max-zaytsev:
Parent issue: [364](https://github.com/microsoft/build-task-team/issues/364)

**Description**

We get the following error during the compilation of XcodeV5 task in _azure-pipelines-tasks_ repo using a newer version of _azure-pipelines-task-lib_:

> Argument of type 'string[]' is not assignable to parameter of type 'string'.

image

`mathcingTestResultsFiles` has `string[]` type [here](https://github.com/microsoft/azure-pipelines-tasks/blob/1538fd6fdb8efd93539b7fe65b00df900d963c1a/Tasks/XcodeV5/postxcode.ts#L38) as it's the result of `findMatch` [function](https://github.com/microsoft/azure-pipelines-task-lib/blob/cf18a238d0ec500d8e8e86b2ff74f901cc92aec2/node/task.ts#L1501). There are several more tasks with the same behavior in the _azure-pipelines-tasks_ .
So we can use `string[]` type (instead of `string`) for the first parameter of  _publish_ method.

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "3.0.4-preview",
+  "version": "3.0.4-preview.0",
   "description": "Azure Pipelines Task SDK",
   "main": "./task.js",
   "typings": "./task.d.ts",
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "3.0.3-preview",
+  "version": "3.0.4-preview",
   "description": "Azure Pipelines Task SDK",
   "main": "./task.js",
   "typings": "./task.d.ts",
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "3.0.3-preview",
+  "version": "3.0.4-preview",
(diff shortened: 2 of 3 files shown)


egor-bryzgalov (approved):

EzzhevNikita (approved):

damccorm (approved):

node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "3.0.3-preview",
+  "version": "3.0.4-preview.0",
```
damccorm, line 3: I think this should just be 3.0.4-preview

damccorm (approved):

node/package.json
max-zaytsev: This was implemented to resolve the issue with task-lib in common packages without using _peerDependencies_.
When we use this pattern _"3.0.3-preview"_ for versioning we are not able to use ^ in the dependencies to allow installing a newer version.
If we implement it this way _"3.0.4-preview.0"_ we would be able to use semantic versioning syntax and resolve the issue with task-lib version conflicts.

damccorm: Ah, I missed the email explaining this - my mistake!

damccorm:
Published!

max-zaytsev:
@damccorm thank you!

PR #356, opened 2018-05-12
Added SecureFile functions in PS SDK

mmajcica:
Added two new input functions in PS SDK Get-VstsSecureFileName and Get-VstsSecureFileTicket to match the available operations in Node SDK. For each new cmdlet, there is also a test associated.

PR #695, opened 2021-01-08
Updated publish methods' to keep it compatible

egor-bryzgalov:
Updated publish methods' to keep it compatible with tasks using different argument types.

Related issue (internal): [#356](https://github.com/microsoft/build-task-team/issues/356)

The approach was chosen taking the following case into consideration:

- If an empty string is passed to `publish` method of `CodeCoveragePublisher` class as `additionalCodeCoverageFiles` argument we would get the following result:
`##vso[codecoverage.publish]`

- If we try to pass an empty array we would get:
`##vso[codecoverage.publish additionalcodecoveragefiles=;]`

The approach is to pass a string in any case since it's expected by the `TaskCommand` class.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -1770,12 +1770,12 @@ export class TestPublisher {
     constructor(public testRunner: string) {
     }
 
-    public publish(resultFiles?: string[], mergeResults?: string, platform?: string, config?: string, runTitle?: string, publishRunAttachments?: string, testRunSystem?: string) {
+    public publish(resultFiles?: string | string[], mergeResults?: string, platform?: string, config?: string, runTitle?: string, publishRunAttachments?: string, testRunSystem?: string) {
         // Could have used an initializer, but wanted to avoid reordering parameters when converting to strict null checks
         // (A parameter cannot both be optional and have an initializer)
         testRunSystem = testRunSystem || "VSTSTask";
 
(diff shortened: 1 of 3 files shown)


node/task.ts
```diff
@@ -1799,7 +1799,7 @@ export class TestPublisher {
         }
 
         if (resultFiles) {
-            properties['resultFiles'] = resultFiles;
+            properties['resultFiles'] = Array.isArray(resultFiles) ? resultFiles.join() : resultFiles;
```
DaniilShmelev, line 1802: Are you sure this won't introduce any unexpected changes?
Correct me if I'm wrong, but seems that with this change [this line in taskcommand](https://github.com/microsoft/azure-pipelines-task-lib/blob/f325a44515b6852b29d5de65af7666e77c84333b/node/taskcommand.ts#L37) will work differently

egor-bryzgalov: This line you've mentioned handles arrays the same way as strings except the case with an empty string or array. This change is introduced to make it handled the same way in any case. If an array is passed it would be 'joined' with a standard separator.

alexander-smolyakov (approved):

DaniilShmelev (approved):

EzzhevNikita (approved):

anatolybolshakov (approved):

PR #717, opened 2021-02-18
[bug fix] Fix unavailable public variables when creating tests

### LINKED ISSUE #716: How can I mock a pipeline variable in my tests?
### Environment
azure-pipelines-task-lib version: 2.9.3

### Issue Description
Hello,

I'd like to mock pipeline variables in my tests, as the way I can configure in the UI below:
image

I'm currently writing my test as the following code (I removed the import and other stuff to focus on the problem): 

**test.ts**:
```

let tmr: tmrm.TaskMockRunner = new tmrm.TaskMockRunner(taskPath);

tmr.setVariableName('MYPREFIX_FOO', 'value_here', false);
tmr.run();

```

**index.ts** (Task code):

```
import tl = require("azure-pipelines-task-lib/task")

async function run() {
  let vars = tl.getVariables()

  console.log(tl.getVariables().length) // it shows 0 instead of 1 regarding the MYPREFIX_FOO variable set in the test.ts when I ran the tests. So I can't access MYPREFIX_FOO at all in the test
}

run()

```

### Expected behaviour

Pipeline variable be available to `tl.getVariables()` 

### Actual behaviour

Variable is not available to the `tl.getVariables()`

I'm not sure if I'm missing something. Am I using the correct command to achieve the mock variable?

Thanks

bartier:
Closes #716

Currently, it's not possible to create public mock variables using the function ````setVariableName('key', 'value', false)```` from a **TaskMockRunner**. This PR attempts to fix this, so it handles the public task variables when ``VariableInfo[]`` should be returned by the function ```getVariables()```.

Example:

Consider this test idea:
````
let tmr: tmrm.TaskMockRunner = new tmrm.TaskMockRunner(taskPath);
tmr.setVariableName('FOO', 'BAR', false); // this FOO variable isn't available when we call  getVariables() in the task
tmr.setVariableName('EGG', 'BAR', true);
tmr.run();
````

If our task needs to access a public variable using ``getVariables``, it only gets the secrets (in this example only EGG).
````
import tl = require("azure-pipelines-task-lib/task")

async function run() {
  let vars = tl.getVariables()

  console.log(tl.getVariables()) // FOO will not be available here, only EGG will be available
}

run()
````

I added a test to validate this case, so I hope this PR could be merged as soon as possible =)

### CODE REVIEW

**Comment by bartier** [comment]
@egor-bryzgalov Could you please have a look at this? I'm not sure if this PR contains the required stuff to be merged :D

**Comment by bartier** [comment]
@damccorm If there is anything I can clarify in the review process I am willing to help =)

File: node/internal.ts [hunk 1]
```diff
@@ -727,7 +733,9 @@ export function _loadData(): void {
                 ++loaded;
                 _debug('loading ' + envvar);
                 _vault.storeSecret(envvar, value);
-                delete process.env[envvar];
+                if (!_startsWith(envvar, 'VSTS_TASKVARIABLE_')) { // avoid deleting public variables
```
anatolybolshakov, line 736: Does it mean that avoiding removal of these variables would expose them for further tasks?

bartier: @anatolybolshakov Good point. I changed the code to continue deleting it to avoid exposing them for further tasks. Can you check it again?

node/internal.ts
```diff
@@ -243,7 +243,11 @@ export function _getVariable(name: string): string | undefined  {
     }
     else {
         // get the public value
-        varval = process.env[key];
+        if (_startsWith(key, 'VSTS_TASKVARIABLE_')) {
+            varval = _knownVariableMap[key].value
```
anatolybolshakov, line 247: Could you please add null check here - so undefined value will be returned for this case, but not an exception being thrown? (the same as for  process.env[key])

anatolybolshakov:
@bartier I'm still not confident about necessity of these changes - why we can't just use `TaskMockRunner.setAnswers` method to provide mock for `GetVariables` task-lib method?

bartier:
@anatolybolshakov Could you please give an example of using setAnswers to achieve that? I'm not sure what field of `TaskLibAnswers` it would be used.

anatolybolshakov:
@bartier oh I see - mock-task does not support mocking for getVariables method at the moment and just [uses](https://github.com/microsoft/azure-pipelines-task-lib/blob/21dac986b4c7478eb2ed8ffbd60e4fad62904438/node/mock-task.ts#L57) original method.
Probably we could extend mock-task - to be able to set up variables - let us take a look at it

Issue #718, opened 2021-02-22
Allow NodeJS binaries to be downloaded from a Mirror URL


ricksbrown:
### Environment

azure-pipelines-task-lib version: All

### Issue Description

When attempting to use `azure-pipelines-task-lib/mock-test` behind a corporate proxy tests fail because of [this line](https://github.com/microsoft/azure-pipelines-task-lib/blob/578d7074e317355d638bea693b562bce8744731a/node/mock-test.ts#L232) which hard codes the nodejs download url:

```js
const nodeUrl: string = 'https://nodejs.org/dist';
```

In highly secure environments these cloud URLs are effectively inaccessible, but on-prem mirrors are available instead.

### Expected behaviour

The SDK should allow developers to configure a mirror URL to use instead of `https://nodejs.org/dist`.

### Actual behaviour

The nodejs download URL is hard-coded.

DaniilShmelev:
Hi @ricksbrown! Thanks for your PR, we'll take a look at it!

DaniilShmelev:
@ricksbrown thanks for your contribution! I've merged your PR, these changes will be released in the next couple of days

PR #719, opened 2021-02-22
Allow mirror URL to be set for NodeJS downloads

ricksbrown:
## Testing

The existing tests fail completely without this change, if you try to run them where nodejs.org is not visible (e.g. behind a corporate proxy).

With this change the tests all pass, assuming you set `TASK_NODE_URL` to a mirror URL.

Closes issue #718

### DIFF
--- a/node/buildutils.js
+++ b/node/buildutils.js
@@ -33,7 +33,8 @@ exports.getExternals = function () {
 
     // download the same version of node used by the agent
     // and add node to the PATH
-    var nodeUrl = 'https://nodejs.org/dist';
+    var nodeUrl = process.env['TASK_NODE_URL'] || 'https://nodejs.org/dist';
+    nodeUrl = nodeUrl.replace(/\/$/, '');  // ensure there is no trailing slash on the base URL
     var nodeVersion = 'v10.23.0';
     switch (platform) {
         case 'darwin':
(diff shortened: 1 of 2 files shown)


node/buildutils.js
```diff
@@ -33,7 +33,8 @@ exports.getExternals = function () {
 
     // download the same version of node used by the agent
     // and add node to the PATH
-    var nodeUrl = 'https://nodejs.org/dist';
+    var nodeUrl = process.env['TASK_NODE_URL'] || 'https://nodejs.org/dist';
+    nodeUrl = nodeUrl.replace(/\/$/, '');  // ensure there is no trailing slash on the base URL
```
anatolybolshakov, line 37: Could you please add null check for nodeUrl?

anatolybolshakov:
@ricksbrown thanks for contribution! Could you please check the comments?

node/buildutils.js
ricksbrown: Hi @anatolybolshakov - As far as I can see it is impossible for `nodeUrl` to be null here. 

Null is already checked here:

`var nodeUrl = process.env['TASK_NODE_URL'] || 'https://nodejs.org/dist';`

If the environment variable is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) (which includes null) then it will default to  `https://nodejs.org/dist`. This happens due to the [behavior](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR) of the `||` operator

Am I missing something?

anatolybolshakov: You're right, sorry 😃 It's fine with default value

DaniilShmelev (approved):
Tested this manually, found no issues

PR #753, opened 2021-04-14
Update to use agent-supported version of escaping

damccorm:
We switched this to %AZP25 in the agent repo and turned it on a month ago - https://github.com/microsoft/azure-pipelines-agent/pull/3303

We should make that switch in the task-lib as well now so that when we encode values they automatically get decoded by the agent


anatolybolshakov (approved):
Thanks, LGTM!

EzzhevNikita (approved):

PR #761, opened 2021-05-10
retry realpathSync

japj:
This is a fix/workaround for https://github.com/microsoft/azure-pipelines-tasks/issues/14789 , where `fs.realpathSync` seems to fail in the first try for a NetApp SnapMirror Volume.

Please note that a new release of `azure-pipelines-task-lib` is required to be integrated in the `azure-pipelines-tasks DeleteFiles task` before the original issue can be fully resolved.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -906,7 +906,14 @@ export function find(findPath: string, options?: FindOptions): string[] {
 
                 if (options.followSymbolicLinks) {
                     // get the realpath
-                    let realPath: string = fs.realpathSync(item.path);
+                    let realPath: string;
+                    try {
+                        realPath = fs.realpathSync(item.path);
+                    } 
+                    catch {
+                        // retry once in case of failure
+                        realPath = fs.realpathSync(item.path);
+                    }
 
                     // fixup the traversal chain to match the item level
                     while (traversalChain.length >= item.level) {


DenisRumyantsev:
Hi @japj, could you please insert this `retry` function before the `find` function?
```node
/**
 * Interface for RetryOptions
 *
 * Contains "continueOnError" and "retryCount" options.
 */
export interface RetryOptions {

    /**
     * If true, code still continues to execute when all retries failed.
     */
    continueOnError: boolean,

    /**
     * Number of retries.
     */
    retryCount: number
}

/**
 * Tries to execute a function a specified number of times.
 *
 * @param   func            a function to be executed.
 * @param   args            executed function arguments array.
 * @param   retryOptions    optional. Defaults to { continueOnError: false, retryCount: 0 }.
 * @returns the same as the usual function
 */
function retry(func: Function, args: any[], retryOptions: RetryOptions = { continueOnError: false, retryCount: 0 }): any {
    while (retryOptions.retryCount >= 0) {
        try {
            return func(...args);
        } catch (e) {
            if (retryOptions.retryCount <= 0) {
                if (retryOptions.continueOnError) {
                    warning(e);
                    break;
                } else {
                    throw e;
                }
            } else {
                console.log(`Attempt to execute function "${func?.name}" failed, retries left: ${retryOptions.retryCount}`);
                retryOptions.retryCount--;
            }
        }
    }
}

/**
 * Recursively finds all paths a given path. Returns an array of paths.
 *
 * @param     findPath  path to search
 * @param     options   optional. defaults to { followSymbolicLinks: true }. following soft links is generally appropriate unless deleting files.
 * @param     realpathSyncRetryOptions  optional. Allows setting retry logic for "fs.realpathSync" method.
 * @returns   string[]
 */
export function find(findPath: string, options?: FindOptions, realpathSyncRetryOptions?: RetryOptions): string[] {
```

And change the line `let realPath: string = fs.realpathSync(item.path);` to:

```node
let realPath: string = retry(fs.realpathSync, [item.path], realpathSyncRetryOptions);
```

After that, we can make an additional `retryCount` input in the `DeleteFiles` task.

japj:
@DenisRumyantsev just to be clear: this will only work if the Azure DevOps Server/Service will also actually set a retryCount on the DeleteFiles task in the DeleteArtifacts job it creates for handling the retention.

Since we have no control on this DeleteArtifacts job creation, we cannot ensure that a retryCount is correctly set. So it will require a server patch to get this working correctly end-to-end (if the retryCount is optional and defaults to 0).

DenisRumyantsev:
Hi @japj, we decided to implement retry logic with default retries number more than zero only for UNC-paths, like the ones you use.

[Related Pull Request](https://github.com/microsoft/azure-pipelines-task-lib/pull/777)

DenisRumyantsev:
Task-lib was updated and the DeleteFiles task awaits deployment.

Issue #667, opened 2020-09-29
Error setting resourceFile: "Maximum call stack size exceeded"


russellseymour:
### Environment
azure-pipelines-task-lib version: 2.10 & 2.11.1

### Issue Description
I am trying to use the `setResourceFile` method to use the `task.json` file of my task. To do this I am using `__dirname` in the TS file, but this is returning `null` or `/` which means that the tasklib is looking for my task in `/task.json` which it cannot find. It then proceeds to error out with `##[error]RangeError: Maximum call stack size exceeded`.

I feel there is something fundamental to this in that it is strange `__dirname` does not have a value. I am also assuming this is something to do with my implementation, but I cannot see what as I am not (as far as I am aware) overwriting the `__dirname` value. Additionally the `__filename` is being set to `index.js` which is not what I was expecting, especially as I do not have an `index.js` in my extension; is this the path to the script that is running my extension? If so how do I get the actual path to my extension file so I can set the resource file?

(I tested running this without any of my code, I commented out my libs and the run function but I still get the same issue).

This is a screen shot of the output from AzDo:

![image](https://user-images.githubusercontent.com/791658/94533098-a4b8dc00-0236-11eb-873a-f187e6973125.png)

### Expected behaviour
I should be able to set the resource file to my `task.json` using the `__dirname` variable.

### Actual behaviour
The `__dirname` var is not being set to the path of my task, which means the tasklib is not able to find the resource file at `/task.json`. I suspect then it goes off in a loop to look for other files, but it then exceeds the number of times it can do this, hence the error.

### Steps to reproduce
1.  The following code shows how I am trying to set the resource file before I run the task

```typescript
// Import the necessary tasks
import * as tl from "azure-pipelines-task-lib"; // task library for Azure DevOps
import * as task from "../common/taskConfiguration";
import * as ex from "../common/executeComponent";
import { join as pathJoin } from "path";
import { sprintf } from "sprintf-js";

// Execute the chosen Chef component command with the specified arguments
async function run() {

  let taskConfiguration = new task.TaskConfiguration();

  // get the parameters for the task, e.g. the settings that have been provided
  await taskConfiguration.getTaskParameters({}, []);

  // create an instance of the executeComponent class
  let executeComponent = new ex.ExecuteComponent(taskConfiguration);

  // perform the execution of the selected command
  executeComponent.Execute();
}

// set the path to the resourceFile
try {
  let resourceFile = pathJoin(__dirname, "task.json");
[...]

PR #773, opened 2021-07-05
 #667 Error setting resourceFile: "Maximum call stack size exceeded"

cima:
+ Preventing the infinite recursion by not translating warning about missing translation.

Please see and reopen issue #667 . What I provide here is fix of catastrphic failure in infinite recursion. It doesn't fix translation in compilation systems like webpack.


node/internal.ts
```diff
@@ -200,10 +200,10 @@ export function _loc(key: string, ...param: any[]): string {
     }
     else {
         if (Object.keys(_resourceFiles).length <= 0) {
-            _warning(_loc('LIB_ResourceFileNotSet', key));
+            _warning(`Resource file haven\\'t been set, can\\'t find loc string for key: ${path}`);
```
anatolybolshakov, line 203:
```suggestion
            _warning(`Resource file haven\'t been set, can\'t find loc string for key: ${path}`);
```

node/internal.ts
```diff
@@ -200,10 +200,10 @@ export function _loc(key: string, ...param: any[]): string {
...
         }
         else {
-            _warning(_loc('LIB_LocStringNotFound', key));
+            _warning(`Can\\'t find loc string for key: ${path}`);
```
anatolybolshakov, line 206:
```suggestion
            _warning(`Can\'t find loc string for key: ${path}`);
```

anatolybolshakov:
@cima thanks for the fix here!
Could you please take a look at suggestions above and bump patch number of package version also?

cima:
@anatolybolshakov I have taken a look at it and I wasn't happy with my own code. I've just add a brief unit test to be able to put breakpoint and debug and discovered two things:

1. Backslash is not necessary at all as it is in backtics that translates to double quotes so the direct apostrophe is not interfering with compiler when parsing a string quotation.
2. The parameter in formated string was totally wrong. It supposed to be `key` not `path`.

I have compile it, run tests and the fix is pushed in the branch.

node/test/internalhelpertests.ts
```diff
@@ -401,4 +401,10 @@ describe('Internal Path Helper Tests', function () {
 
         done();
     });
+    
+    it('ReportMissingLocalization', (done: MochaDone) => {
```
anatolybolshakov, line 405: @cima this test does not really new changes - it passes successfully with and without new changes.

cima: You are right that the test finishes OK regardles off my change. The propper testing would require some serious mocking using rewire mock, which would require new dependencies. Maybe damaging `__dirname` variable within the unittest might do the trick. It is bit strettching beyond replacing two strings. Shall I fully simulate the situation when no file is loaded?

anatolybolshakov: @cima yes, I guess that would be more helpful for coverage purposes.

cima: I have included a test that simulates the missing file. When such test (5da2a2a98428d6300bde4e529237ad181c1bd2a1) is applied to master (ba1f94be19f958f1f718ac946aef3123c8eee4a8) it triggers the infinite recusion.

anatolybolshakov: @cima thanks a lot!

anatolybolshakov (approved):
LGTM, thanks for contribution!

AnnaOpareva (approved):

PR #774, opened 2021-07-09
Skipping the file if it was not found

DenisRumyantsev:
**Description**: occasionally the `find` function fails with the error `Failed find: ENOENT: no such file or directory`. The reason is that the `fs.lstatSync` method applies to a file that already was deleted. So we should check a file presence and skip it if it does not exist.

**Changelog**: `node/task.ts` — the `find` function — checking a file presence added.

[Related issue](https://github.com/microsoft/azure-pipelines-tasks/issues/14816)

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -897,6 +897,10 @@ export function find(findPath: string, options?: FindOptions): string[] {
                 }
             }
             else {
+                if (!fs.existsSync(item.path)) {
+                    debug(`File ${item.path} has not been found and will be ignored.`);
+                    continue;
+                }
                 // use lstat (not following symlinks)
                 stats = fs.lstatSync(item.path);
             }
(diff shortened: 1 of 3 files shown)


anatolybolshakov:
Please bump package version also

node/task.ts
```diff
@@ -897,6 +897,10 @@ export function find(findPath: string, options?: FindOptions): string[] {
                 }
             }
             else {
+                if (!fs.existsSync(item.path)) {
+                    debug(`File "${item.path}" has not been found and will be skipped.`);
```
anatolybolshakov, line 901: Could it make sense to change it to something like 'File "${item.path}" seems to be removed during find operation execution - so skipping it.'?
Currently it is unclear why do we mention some not existing path in logs.

anatolybolshakov (approved):
LGTM, thanks!

alexander-smolyakov (approved):

EzzhevNikita (approved):

PR #779, opened 2021-07-19
Skipping the file if it was not found (the CI is fixed)

DenisRumyantsev:
**Description**: occasionally the `find` function fails with the error `Failed find: ENOENT: no such file or directory`. The reason is that the `fs.lstatSync` method applies to a file that already was deleted. So we should check a file presence and skip it if it does not exist.

**Changelog**: `node/task.ts` — the `find` function — a part of the code where the error occurs was taken out to the separate `_getStats` function. A new field `skipMissingFiles` was added into the `findOptions` parameter. When it is true, the `find` function skips files that are not found during the `_getStats` function execution.

[Related issue](https://github.com/microsoft/azure-pipelines-tasks/issues/14816)

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -857,11 +857,11 @@ export function find(findPath: string, options?: FindOptions): string[] {
         while (stack.length) {
             // pop the next item and push to the result array
             let item = stack.pop()!; // non-null because `stack.length` was truthy
+            result.push(item.path);
             if (!fs.existsSync(item.path)) {
                 debug(`File "${item.path}" seems to be removed during find operation execution - so skipping it.`);
                 continue;
             }
-            result.push(item.path);
 
             // stat the item.  the stat info is used further below to determine whether to traverse deeper
             //
--- a/node/task.ts
+++ b/node/task.ts
@@ -903,6 +903,10 @@ export function find(findPath: string, options?: FindOptions): string[] {
         while (stack.length) {
             // pop the next item and push to the result array
             let item = stack.pop()!; // non-null because `stack.length` was truthy
+            if (!fs.existsSync(item.path)) {
+                debug(`File "${item.path}" seems to be removed during find operation execution - so skipping it.`);
+                continue;
+            }
             result.push(item.path);
 
             // stat the item.  the stat info is used further below to determine whether to traverse deeper
@@ -918,11 +922,6 @@ export function find(findPath: string, options?: FindOptions): string[] {
                 catch (err) {
                     if (err.code == 'ENOENT' && options.allowBrokenSymbolicLinks) {
                         // fallback to lstat (broken symlinks allowed)
-                        if (!fs.existsSync(item.path)) {
-                            debug(`File "${item.path}" seems to be removed during find operation execution - so skipping it.`);
-                            result.pop();
-                            continue;
-                        }
                         stats = fs.lstatSync(item.path);
                         debug(`  ${item.path} (broken symlink)`);
                     }
--- a/node/task.ts
+++ b/node/task.ts
@@ -920,13 +920,13 @@ export function find(findPath: string, options?: FindOptions): string[] {
                         // fallback to lstat (broken symlinks allowed)
                         try {
                             stats = fs.lstatSync(item.path);
-                        } catch (e) {
-                            if (e.code !== 'ENOENT') {
-                                throw e;
+                        } catch (err2) {
+                            if (err2.code == 'ENOENT') {
+                                debug(`File "${item.path}" seems to be removed during find operation execution - so skipping it.`);
+                                result.pop();
+                                continue;
                             }
-                            debug(`File "${item.path}" seems to be removed during find operation execution - so skipping it.`);
-                            result.pop();
-                            continue;
+                            throw err2;
                         }
                         debug(`  ${item.path} (broken symlink)`);
                     }
--- a/node/task.ts
+++ b/node/task.ts
@@ -1026,13 +1026,15 @@ function _debugFindOptions(options: FindOptions): void {
     debug(`findOptions.allowBrokenSymbolicLinks: '${options.allowBrokenSymbolicLinks}'`);
     debug(`findOptions.followSpecifiedSymbolicLink: '${options.followSpecifiedSymbolicLink}'`);
     debug(`findOptions.followSymbolicLinks: '${options.followSymbolicLinks}'`);
+    debug(`findOptions.skipMissingFiles: '${options.skipMissingFiles}'`);
 }
 
 function _getDefaultFindOptions(): FindOptions {
     return <FindOptions>{
         allowBrokenSymbolicLinks: false,
         followSpecifiedSymbolicLink: true,
-        followSymbolicLinks: true
+        followSymbolicLinks: true,
+        skipMissingFiles: false
     };
 }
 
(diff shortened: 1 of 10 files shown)


node/task.ts
```diff
@@ -858,6 +863,45 @@ export function retry(func: Function, args: any[], retryOptions: RetryOptions =
     }
 }
 
+function _getStats (path: string, isFirst: boolean, options: FindOptions): fs.Stats {
```
max-zaytsev, line 866: It's not clear what is `isFirst `here?

node/task.ts
```diff
@@ -903,49 +947,19 @@ export function find(findPath: string, options?: FindOptions): string[] {
...
-                    if (err.code == 'ENOENT' && options.allowBrokenSymbolicLinks) {
-                        // fallback to lstat (broken symlinks allowed)
-                        stats = fs.lstatSync(item.path);
-                        debug(`  ${item.path} (broken symlink)`);
-                    }
-                    else {
-                        throw err;
-                    }
+            try {
+                stats = _getStats(item.path, !result.length, options);
```
alexander-smolyakov, line 954: Could you please provide some comments regarding `!result.length` value? Also, I think it makes sense to extract this value as a variable and give it an appropriate name, so the code will be easier to read.

EzzhevNikita (commented):
@DenisRumyantsev As far as I understand these changes should fix CI, but looks the issue still persists.

DenisRumyantsev:
> @DenisRumyantsev As far as I understand these changes should fix CI, but looks the issue still persists.

@EzzhevNikita I have already tried to provide similar changes related to [the same issue](https://github.com/microsoft/azure-pipelines-tasks/issues/14816) via [this PR](https://github.com/microsoft/azure-pipelines-task-lib/pull/774), but it broke the CI for Linux and macOS, so I had to [revert](https://github.com/microsoft/azure-pipelines-task-lib/pull/778) it and made a new PR, the current one. For fixing the CI for Windows, there was created [another separate issue](https://github.com/microsoft/build-task-team/issues/1393).

node/task.ts
```diff
@@ -858,6 +863,53 @@ export function retry(func: Function, args: any[], retryOptions: RetryOptions =
     }
 }
 
+/**
+ * Gets info about item stats.
+ *
+ * @param path              a path to the item to be processed.
+ * @param isPathToSearch    true when a path to the item is a path to search.
```
alexander-smolyakov, line 870:
```suggestion
 * @param isPathToSearch    true when a path to the item equals the path to search.
```

alexander-smolyakov (approved):
LGTM and please take a look at the suggestion

EzzhevNikita (approved):

max-zaytsev (approved):

node/task.ts
```diff
@@ -903,49 +942,25 @@ export function find(findPath: string, options?: FindOptions): string[] {
...
+                const isPathToSearch = !result.length;
+
+                // following all symlinks OR following symlinks for the specified path AND this is the specified path
+                const followSymbolicLinks = options.followSymbolicLinks || options.followSpecifiedSymbolicLink && isPathToSearch;
+
+                // stat the item. The stat info is used further below to determine whether to traverse deeper
+                stats = _getStats(item.path, followSymbolicLinks, options.allowBrokenSymbolicLinks);
+            } catch (err) {
+                if (err.code == 'ENOENT' && options.skipMissingFiles) {
+                    debug(`File "${item.path}" seems to be removed during find operation execution - so skipping it.`);
```
max-zaytsev, line 958: We have to fix this message since the error here can be not only due to the missing file but due to a broken symlink also

max-zaytsev (approved):

node/task.ts
```diff
@@ -903,49 +942,25 @@ export function find(findPath: string, options?: FindOptions): string[] {
...
-                        // fallback to lstat (broken symlinks allowed)
-                        stats = fs.lstatSync(item.path);
-                        debug(`  ${item.path} (broken symlink)`);
-                    }
-                    else {
-                        throw err;
-                    }
+            try {
+                // `item.path` equals `findPath` for the first item to be processed, when the `result` array is empty
+                const isPathToSearch = !result.length;
```
alexander-smolyakov, line 949:
```suggestion
                const isPathToSearch: boolean = !result.length;
```

node/task.ts
```diff
@@ -903,49 +942,25 @@ export function find(findPath: string, options?: FindOptions): string[] {
...
+
+                // following all symlinks OR following symlinks for the specified path AND this is the specified path
+                const followSymbolicLinks = options.followSymbolicLinks || options.followSpecifiedSymbolicLink && isPathToSearch;
+
+                // stat the item. The stat info is used further below to determine whether to traverse deeper
+                stats = _getStats(item.path, followSymbolicLinks, options.allowBrokenSymbolicLinks);
```
alexander-smolyakov, line 955:
```suggestion
                // following specified symlinks only if current path equals specified path 
                const followSpecifiedSymbolicLink: boolean = options.followSpecifiedSymbolicLink && isPathToSearch;
                
                // following all symlinks or following symlinks for the specified path
                const allowFollowSymbolicLink: boolean = options.followSymbolicLinks || followSpecifiedSymbolicLink;
                
                // stat the item. The stat info is used further below to determine whether to traverse deeper
                stats = _getStats(item.path, allowFollowSymbolicLink, options.allowBrokenSymbolicLinks);
```

node/task.ts
```diff
@@ -858,6 +863,40 @@ export function retry(func: Function, args: any[], retryOptions: RetryOptions =
...
 
+/**
+ * Gets info about item stats.
+ *
+ * @param path                      a path to the item to be processed.
+ * @param followSymbolicLinks       indicates whether to traverse descendants of symbolic link directories.
+ * @param allowBrokenSymbolicLinks  when true, broken symbolic link will not cause an error.
+ * @returns fs.Stats
+ */
+function _getStats (path: string, followSymbolicLinks: boolean, allowBrokenSymbolicLinks: boolean): fs.Stats {
```
alexander-smolyakov, line 874:
```suggestion
function _getStats (path: string, allowFollowSymbolicLinks: boolean, allowBrokenSymbolicLinks: boolean): fs.Stats {
```

node/task.ts
```diff
@@ -903,49 +942,25 @@ export function find(findPath: string, options?: FindOptions): string[] {
...
+                const isPathToSearch = !result.length;
+
+                // following all symlinks OR following symlinks for the specified path AND this is the specified path
+                const followSymbolicLinks = options.followSymbolicLinks || options.followSpecifiedSymbolicLink && isPathToSearch;
+
+                // stat the item. The stat info is used further below to determine whether to traverse deeper
+                stats = _getStats(item.path, followSymbolicLinks, options.allowBrokenSymbolicLinks);
+            } catch (err) {
+                if (err.code == 'ENOENT' && options.skipMissingFiles) {
+                    warning(`"${item.path}" seems to be a removed file or directory / broken symlink - so skipping it.`);
```
alexander-smolyakov, line 958: Is it possible to more clearly define the root cause of the error? And display a more accurate error message?

DenisRumyantsev: Have no ideas how to catch the difference between a removed file or directory and a broken symlink.
Just as it is impossible to say exactly how and when a file or directory was removed.

Issue #621, opened 2020-03-05
Unable to mock Windows in `getPlatform` call


tillig:
### Environment
azure-pipelines-task-lib version: 2.9.3

### Issue Description
When mocking `getPlatform` you can't set a response that indicates Windows.

The mock answers handler returns `null` [if the value of the mock answer isn't truthy](https://github.com/microsoft/azure-pipelines-task-lib/blob/34e17f6ca02c256bcb4e121880f6d4b7da65eed2/node/mock-answer.ts#L45). The [`Platform` enum has `Windows` as the first entry](https://github.com/microsoft/azure-pipelines-task-lib/blob/34e17f6ca02c256bcb4e121880f6d4b7da65eed2/node/task.ts#L49) which means the value of `Platform.Windows` is zero.

Here's the rendered JS:

```js
/** Platforms supported by our build agent */
var Platform;
(function (Platform) {
    Platform[Platform["Windows"] = 0] = "Windows";
    Platform[Platform["MacOS"] = 1] = "MacOS";
    Platform[Platform["Linux"] = 2] = "Linux";
})(Platform = exports.Platform || (exports.Platform = {}));
```

What that means is if you have answers defined:

```typescript
const answers: ma.TaskLibAnswers = {
  'getPlatform': {
    'getPlatform': tl.Platform.Windows,
  },
};
runner.setAnswers(answers);
```

Then you'll see something like this in your debug output:

```text
##vso[task.debug]looking up mock answers for "getPlatform", key '"getPlatform"'
##vso[task.debug]mock response not found
```

This becomes troublesome for code like this:

```typescript
export function setConsoleCodePage(): void {
  if (tl.getPlatform() === tl.Platform.Windows) {
    tl.execSync(path.resolve(process.env.windir as string, 'system32', 'chcp.com'), ['65001']);
  }
}
```

You can work around this by being slightly less explicit:

```typescript
export function setConsoleCodePage(): void {
  if (!tl.getPlatform()) {
    tl.execSync(path.resolve(process.env.windir as string, 'system32', 'chcp.com'), ['65001']);
  }
}
```

but it'd be nice if the mocking didn't drop non-truthy answers.

[...]

PR #788, opened 2021-08-19
Change implicit check in getResponse to explicit

kuleshovilya:
Related to [issue](https://github.com/microsoft/azure-pipelines-task-lib/issues/621) 
Changed check for null to explicit comparison to avoid zero number being checked as null.

### DIFF
--- a/node/mock-answer.ts
+++ b/node/mock-answer.ts
@@ -47,7 +47,8 @@ export class MockAnswers {
 
         const cmd_answer = this._answers[cmd]!;
 
-        if (cmd_answer[key]) {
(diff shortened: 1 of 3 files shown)


anatolybolshakov:
@kuleshovilya should we bump package version also?

kuleshovilya:
@anatolybolshakov that's the question we are thinking of, whether it's needed or not? Functionality itself doesn't change, its only a minor change for mock module. So I'm thinking no?

anatolybolshakov (approved):

EzzhevNikita (approved):

Issue #775, opened 2021-07-09
`getEndpointAuthorization` does not throw when authorization was not set and is not optional


SimonAlling:
### Environment
azure-pipelines-task-lib version: 3.1.4

### Issue Description
Excerpt from the JSDoc comment about [`getEndpointAuthorization` in `node/task.ts`](../blob/bbec8ff7dc74d50a3aecd643fd1e899f3baf3cb6/node/task.ts#L416):

> If the authorization was not set and is not optional, it will throw.

I believe this is incorrect.

### Expected behaviour

The function throws if the authorization was not set and is not optional.

### Actual behaviour

The function does not throw.

### Steps to reproduce

```typescript
import * as tl from "azure-pipelines-task-lib/task"

try {
  console.log("VALUE:", tl.getEndpointAuthorization("NONEXISTENT", false))
} catch (err) {
  console.warn("CAUGHT:")
  console.warn(err)
}
```

Output when run on my local machine:

```
##vso[task.debug]agent.TempDirectory=undefined
##vso[task.debug]agent.workFolder=undefined
##vso[task.debug]loading inputs and endpoints
##vso[task.debug]loaded 0
##vso[task.debug]Agent.ProxyUrl=undefined
##vso[task.debug]Agent.CAInfo=undefined
##vso[task.debug]Agent.ClientCert=undefined
##vso[task.debug]Agent.SkipCertValidation=undefined
##vso[task.debug]task result: Failed
##vso[task.issue type=error;]Endpoint auth data not present: NONEXISTENT
##vso[task.complete result=Failed;]Endpoint auth data not present: NONEXISTENT
##vso[task.debug]NONEXISTENT exists true
VALUE: undefined
```

Note that it also says `NONEXISTENT exists true`. In fact, it _always_ does that, because [the expression being logged](../blob/bbec8ff7dc74d50a3aecd643fd1e899f3baf3cb6/node/task.ts#L423) is

```typescript
id + ' exists ' + (aval !== null)
```

and [`aval` has type `string | undefined`](../blob/bbec8ff7dc74d50a3aecd643fd1e899f3baf3cb6/node/task.ts#L417), so its value will probably never be `null`.

SimonAlling:
Any updates on this?

[...]

PR #790, opened 2021-08-31
Fix documentation, comment and condition for getEndpointAuthorization function

AnnaOpareva:
Fix documentation, comment and condition for getEndpointAuthorization function

The condition was changed because **aval** has type string | undefined, so its value will probably never be null and it will be better to use !!aval for more strong verification, it will allow us to check the empty string also.

related issue #775

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "3.1.8",
+  "version": "3.1.9",
   "description": "Azure Pipelines Task SDK",
   "main": "./task.js",
   "typings": "./task.d.ts",
--- a/node/task.ts
+++ b/node/task.ts
@@ -407,7 +407,7 @@ export interface EndpointAuthorization {
 
 /**
  * Gets the authorization details for a service endpoint
- * If the authorization was not set and is not optional, it will write logs and return undefined.
+ * If the authorization was not set and is not optional, it will specify task result as failed in case of exception.
  *
  * @param     id        name of the service endpoint
  * @param     optional  whether the url is optional
@@ -420,7 +420,7 @@ export function getEndpointAuthorization(id: string, optional: boolean): Endpoin
         setResult(TaskResult.Failed, loc('LIB_EndpointAuthNotExist', id));
     }
 
-    debug(id + ' exists ' + (aval !== undefined));
+    debug(id + ' exists ' + (!!aval));
 
     var auth: EndpointAuthorization | undefined;
     try {
--- a/node/task.ts
+++ b/node/task.ts
@@ -407,7 +407,7 @@ export interface EndpointAuthorization {
 
 /**
  * Gets the authorization details for a service endpoint
- * If the authorization was not set and is not optional, it will specify task result as failed in case of exception.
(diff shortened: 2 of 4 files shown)


anatolybolshakov:
Task lib version is not bumped

node/task.ts
```diff
@@ -407,7 +407,7 @@ export interface EndpointAuthorization {
 
 /**
  * Gets the authorization details for a service endpoint
- * If the authorization was not set and is not optional, it will throw.
+ * If the authorization was not set and is not optional, it will write logs and return undefined.
```
anatolybolshakov, line 410: Could you please elaborate more? Will in change task result also? I don't think we need to mention undefined result

tatyana-kostromskaya (approved):

anatolybolshakov (changes requested):
Thanks! Please take a look at comments

node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "3.1.8",
+  "version": "3.1.9",
```
anatolybolshakov, line 3: Should we update package-lock also?

SimonAlling: I think so; `npm version patch` updates both `package.json` and `package-lock.json`.

node/task.ts
```diff
@@ -420,7 +420,7 @@ export function getEndpointAuthorization(id: string, optional: boolean): Endpoin
         setResult(TaskResult.Failed, loc('LIB_EndpointAuthNotExist', id));
     }
 
-    debug(id + ' exists ' + (aval !== null));
+    debug(id + ' exists ' + (!!aval));
```
SimonAlling, line 423: I suggest `aval !== undefined` instead of `!!aval`, because I think that's closer to the intended meaning of the code and has less surprising semantics: for example, `"" !== undefined` is `true`, but `!!""` is `false`.

Also, it might be worth mentioning in the PR description/commit message body _why_ this change is made (namely, because `aval` has type `string | undefined` and thus can't be `null`).

node/task.ts
```diff
@@ -407,7 +407,7 @@ export interface EndpointAuthorization {
 
 /**
  * Gets the authorization details for a service endpoint
- * If the authorization was not set and is not optional, it will throw.
+ * If the authorization was not set and is not optional, it will specify task result as failed in case of exception.
```
SimonAlling, line 410: If I weren't already involved in this part of the code base, I wouldn't have been able to figure out what "specify task result as failed in case of exception" meant without reading the implementation. In particular, I would be confused by "specify" and "in case of exception". Suggestion:

```suggestion
 * If the authorization was not set and is not optional, it will set the task result to Failed.
```

Note that, in `setResult`'s JSDoc comment, that function is described as _setting_ (not specifying) the task result, and "Failed" is capitalized.

SimonAlling (changes requested):

anatolybolshakov (approved):

node/task.ts
AnnaOpareva: Thanks for the detailed comment. I think, in our case, it will be better to use !!aval for more strong verification, it will allow us to check the empty string also.

PR #799, opened 2021-10-15
Add platform FreeBSD

joperator:
Since [PR 3266](https://github.com/microsoft/azure-pipelines-agent/pull/3266) makes it possible to run the Azure Pipelines Agent on FreeBSD, it would also make sense that the Azure Pipelines Task SDK supports this platform as well. Currently, for example, the [Extract Files task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/extract-files) fails on FreeBSD with the following error:

```
##[error]Unhandled: Platform not supported: freebsd
```

This is because [extractfilestask.ts](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/ExtractFilesV1/extractfilestask.ts#L17) calls `getPlatform` to determine whether running on Windows. If the current platform is neither Windows, nor MacOS, nor Linux, `getPlatform` throws the above exception. However, this can be easily fixed with this PR by adding `FreeBSD` to the `Platform` enum and returning it in the `getPlatform` function instead of throwing an exception.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -56,7 +56,8 @@ export const IssueSource = im.IssueSource;
 export enum Platform {
     Windows,
     MacOS,
-    Linux
+    Linux,
+    FreeBSD
 }
 
 export enum AgentHostedMode {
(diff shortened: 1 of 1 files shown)


arrowd:
This is a trivial PR, can it be merged in?

sec:
Any move on this one? :)

littleq0903:
let me try to get someone internally to merge this, stay tuned.

Thefrank:
@littleq0903 any news? any way for the FreeBSD community to help with this?

PR #804, opened 2021-11-23
Add OneLocBuild removal to the localization pipeline

kuleshovilya:
Related [issue](https://github.com/microsoft/build-task-team/issues/1844)
Added removal of the OneLocBuild to the command that removes Localization folder, as well as a parameter to ignore error if the folder doesn't exist, that allows deleting any of the folders that exist while skipping error if one of them doesn't

### DIFF
--- a/Localize/localize-pipeline.yml
+++ b/Localize/localize-pipeline.yml
@@ -71,8 +71,14 @@ stages:
 
         git checkout -b $updateBranch
 
-        Remove-Item -Recurse -Force Localize, OneLocBuild -ErrorAction Ignore
-
+        if (Test-Path -Path Localize) { 
+          Remove-Item -Recurse -Force Localize 
+        }
+        
+        if (Test-Path -Path OneLocBuild) { 
+          Remove-Item -Recurse -Force OneLocBuild 
+        }
+        
         git add -A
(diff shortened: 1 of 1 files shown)


Localize/localize-pipeline.yml
```diff
@@ -71,10 +71,10 @@ stages:
 
         git checkout -b $updateBranch
 
-        Remove-Item -Recurse -Force Localize
+        Remove-Item -Recurse -Force Localize, OneLocBuild -ErrorAction Ignore
```
anatolybolshakov, line 74: Why do we need -ErrorAction Ignore here? What happens if there will be any errors?

kuleshovilya: Well, from the explanation and previous PRs I've gathered that OneLocBuild might not exist, so that way we just ignore if this doesn't exist.
An error will appear pretty much only in case one of the folders doesn't exist(or both) and will delete the one that exists

kuleshovilya: Anyways, fixed to double if for folder checks

max-zaytsev (approved):

anatolybolshakov (approved):

PR #815, opened 2022-02-10
Resolve vulnerabilities in dependencies

AndreyIvanov42:
- Bumped shelljs to 0.8.5
- adm-zip to 0.5.9
- fixed tests (due to changes in methods shelljs)

Tested only with mocha tests


anatolybolshakov:
@AndreyIvanov42 where are ensureStarted and ensureTool methods affected - could you please double-check that you've tested scenarios in which these methods are being used?

AndreyIvanov42:
@anatolybolshakov 
I double-checked `ensureStarted` and `ensureTool` methods. They are used only in tests. 
But since many methods of this library are wrappers over shelljs methods, I need to make sure that backward compatibility is not broken.

anatolybolshakov:
Please bump package/package-lock versions

anatolybolshakov (approved):

max-zaytsev (approved):

kriti218:
@AndreyIvanov42 When are we going to have 3.2.0 as the latest version to download? It looks like shelljs vulnerability was fixed in this PR #815 when version was updated to 0.8.5 from 8.4.

AndreyIvanov42:
@kriti218 
New version is available

Issue #824, opened 2022-03-18
Required inputs undefined. Documentation does not match library.


Banner-Keith:
Please check our current Issues to see if someone already reported this https://github.com/Microsoft/azure-pipelines-task-lib/issues

## Environment
azure-pipelines-task-lib version: 3.2.0

## Issue Description
Two issues

### 1. Documentation out-of-date
The [documentation](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/docs/azure-pipelines-task-lib.md#input-functions) for the input functions is out of date. It shows that they return a string, when they actually return `string | undefined`

### 2. Request for improved API functionality
It is very ungainly and difficult to work with this api because we have to handle the undefined case every time we get a required input. The library already throws an error if the input is not provided, so checking again is not necessary.

```TypeScript
export function getInput(name: string, required?: boolean): string | undefined {
    var inval = im._vault.retrieveSecret('INPUT_' + im._getVariableKey(name));

    if (required && !inval) {
        throw new Error(loc('LIB_InputRequired', name));
    }

    debug(name + '=' + inval);
    return inval;
}
```

It would be nice if we could get a version of each of these methods like this so we don't have to check for undefined on every input we pull in. The existing methods would be unchanged so this change would be backwards compatible.

```TypeScript
// Name could be getRequiredInput if that is preferrable
export function getInputRequired(name: string): string {
    var inval = im._vault.retrieveSecret('INPUT_' + im._getVariableKey(name));

    if (!inval) {
        throw new Error(loc('LIB_InputRequired', name));
    }

    debug(name + '=' + inval);
    return inval;
}
```

#### Example Code

Current behavior:
```TypeScript
const exampleInput: string | undefined = tl.getInput('ExampleInput');
const exampleRequiredInput: string | undefined = tl.getInput('ExampleRequiredInput', true);

if (exampleRequiredInput === undefined) {
    throw new Error("Missing ExampleRequiredInput");
}
```

Desired behavior:
```TypeScript
const exampleInput: string | undefined = tl.getInput('ExampleInput');
const exampleRequiredInput: string = tl.getInputRequired('ExampleRequiredInput');

// No checking required because the required input is never undefined
```

I would be happy to open a Pull Request with these new methods if you would be interested in reviewing my proposal.

anatolybolshakov:
[...]

PR #827, opened 2022-03-22
Added new required functions

Banner-Keith:
Return the value instead of the value or undefined.

getInputRequired
getPathInputRequired
getEndpointUrlRequired
getEndpointDataParameterRequired
getEndpointAuthorizationSchemeRequired
getEndpointAuthorizationParameterRequired

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -236,14 +236,7 @@ export function getInput(name: string, required?: boolean): string | undefined {
  * @returns   string
  */
 export function getInputRequired(name: string): string {
-    var inval = im._vault.retrieveSecret('INPUT_' + im._getVariableKey(name));
-
-    if (!inval) {
-        throw new Error(loc('LIB_InputRequired', name));
-    }
-
-    debug(name + '=' + inval);
-    return inval;
+    return getInput(name, true)!;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


Banner-Keith:
Fixes #824 

I also saw that #826 was opened. I can rebase after that PR is completed.

I have added unit tests for everything I added and all tests are passing.

Banner-Keith:
`getInputRequired` and `getPathInputRequired` are the important ones but I did the same addition on some other similar methods. If you only want some of those I can remove some of the additions.

I can also adjust naming or anything else if there's a suggestion.

node/task.ts
```diff
@@ -228,6 +228,24 @@ export function getInput(name: string, required?: boolean): string | undefined {
...
 
+/**
+ * Gets the value of an input.
+ * If the value is not set, it will throw.
+ *
+ * @param     name     name of the input to get
+ * @returns   string
+ */
+export function getInputRequired(name: string): string {
+    var inval = im._vault.retrieveSecret('INPUT_' + im._getVariableKey(name));
```
anatolybolshakov, line 239: Please use the call of existing methods instead (like getInput(name, required) - to avoid duplication

anatolybolshakov:
Thanks for contribution! Could you please take a look at comments?

node/task.ts
Banner-Keith: Do you mean that I should have `getInputRequired` call `getInput` so that the internals of that method are not duplicated?

anatolybolshakov: @Banner-Keith yep correct 👍

SvetlanaMaliugina (approved):
Tested 'getInputRequired' and 'getPathInputRequired' methods manually. Looks fine for me.

anatolybolshakov:
@mmrazik could you please take a look also?

PR #834, opened 2022-04-28
added formatted proxy URL

kirill-ivlev:
Added full proxy URL (http://user:password@host) for the getHttpProxyMethod 
And cover this method with the unit tests.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -1841,7 +1841,8 @@ export interface ProxyConfiguration {
     proxyUrl: string;
     /**
      * Proxy URI formated as: protocol://username:password@hostname:port
-     * For tools that requires to set proxy configuration in the single environment variable
+     * 
+     * For tools that require setting proxy configuration in the single environment variable
      */
     proxyFormattedUrl: string;
     proxyUsername?: string;
--- a/node/task.ts
+++ b/node/task.ts
@@ -1850,7 +1850,13 @@ export interface ProxyConfiguration {
     proxyBypassHosts?: string[];
 }
 
-
+/**
+ * Build Proxy URL in the following format: protocol://username:password@hostname:port
+ * @param proxyUrl Url address of the proxy server (eg: http://example.com)
+ * @param proxyUsername Proxy username (optional)
+ * @param proxyPassword Proxy password (optional)
+ * @returns string
+ */
 function getProxyFormattedUrl(proxyUrl: string, proxyUsername: string | undefined, proxyPassword: string | undefined): string {
     const parsedUrl: URL = new URL(proxyUrl);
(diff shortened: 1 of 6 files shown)


kirill-ivlev:
@mmrazik Could you please review this PR

node/task.ts
```diff
@@ -1797,11 +1798,18 @@ export function getHttpProxyConfiguration(requestUrl?: string): ProxyConfigurati
             return null;
         }
         else {
+            const parsedUrl: URL = new URL(proxyUrl);
+            let proxyAddress: string = `${parsedUrl.protocol}//${parsedUrl.host}`;
+            if (proxyUsername) {
+                proxyAddress = `${parsedUrl.protocol}//${proxyUsername}:${proxyPassword}@${parsedUrl.host}`;
+            }
+            console.log('returning proxy config');
```
anatolybolshakov, line 1806: Do we really need to signalize about it in stdout (not even on debug level)?

node/task.ts
```diff
@@ -1767,6 +1767,7 @@ export function findMatch(defaultRoot: string, patterns: string[] | string, find
 
 export interface ProxyConfiguration {
     proxyUrl: string;
+    proxyFormattedUrl: string;
```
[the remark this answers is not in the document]
kirill-ivlev: JSDoc added
Docs updated

node/task.ts
kirill-ivlev: removed

node/task.ts
```diff
@@ -1839,11 +1839,27 @@ export function findMatch(defaultRoot: string, patterns: string[] | string, find
...
+     * For tools that require setting proxy configuration in the single environment variable
+     */
+    proxyFormattedUrl: string;
     proxyUsername?: string;
     proxyPassword?: string;
     proxyBypassHosts?: string[];
 }
 
+
+function getProxyFormattedUrl(proxyUrl: string, proxyUsername: string | undefined, proxyPassword: string | undefined): string {
```
anatolybolshakov, line 1854: Please add js doc to this new method as well

anatolybolshakov:
Could you please bump package version as well (patch number)?

anatolybolshakov:
LGTM, thanks! Please take a look at the comments also

anatolybolshakov (approved):

node/task.ts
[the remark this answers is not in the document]
anatolybolshakov: Resolved per offline discussion - since we considering that this is intended to be read-only object (containing current proxy settings) - it should be good to keep in a separate field

alexander-smolyakov (approved):

alexander-smolyakov:
@mmrazik would you mind taking a look at this pull request?

mmrazik (approved):

PR #922, opened 2023-03-14
Update toolrunner.ts

andrelvahter:
In case the command line arguments contain secrets or any other info that should not be printed out. We could use the same 'silent' flag to not add all the arguments to debug log.


andrelvahter:
> @microsoft-github-policy-service agree company="Microsoft"

@microsoft-github-policy-service agree company="Microsoft"

Issue #955, opened 2023-07-18
CVE-2022-37614: Critical Vulnerability in mockery module


luxaflow:
Please check our current Issues to see if someone already reported this https://github.com/Microsoft/azure-pipelines-task-lib/issues

### Environment
azure-pipelines-task-lib version: 4.4.0

### Issue Description
These is a critical vulnerability in mfncooper/mocker 2.1.0 module used.

### Expected behaviour
No issues when used, but also no critical security CVE's in used in modules

### Actual behaviour
Currently critical security CVE in mfncooper/mockery module
https://nvd.nist.gov/vuln/detail/CVE-2022-37614

### Steps to reproduce
N/A

### Logs
Link to CVE: https://nvd.nist.gov/vuln/detail/CVE-2022-37614

aleksandrlevochkin:
Hi @luxaflow thank you for reporting this issue, this is currently being worked on.

joshftb:
@maksimu mockery is only used for testing, so none of its exports will see prod. Also, the task does not use the vulnerable component of that library anyways.

Vertex-btb:
@joshftb Two weeks ago this was given the 'triage' label and then removed. Your latest response indicates that this vulnerability may not be exploitable, yet the issue remains 'Open.'  Please advise if Microsoft is planning to update their references to use the newest version of mockery that is not vulnerable _([according to SNYK there are no 'next non-vulnerable versions'](https://security.snyk.io/vuln/SNYK-JS-MOCKERY-3043117))_ or plans to close this issue.

rajarajan2801:
Team, Any updates on this? If the task doesn't use the vulnerable component, Can you please advise why this issue is not closed? We need to respond back to our Security team about the status of this issue.

aleksandrlevochkin:
@Vertex-btb, @rajarajan2801 Mockery is indeed used only for testing. This issue is not closed, as we're preparing a replacement, and it's not merged yet.

JuanDuhalde12:
Hi, any updates on this?. I have a task created and I depends on solve this issue to be able to deploy it. Thanks

lgmorand:
hi :)
not a single commit on the PR for a full month since the Lilia's work. any idea when we could expect it or what is blocking because all checks seem OK ? thanks <3

jessehouwing:
[...]

PR #961, opened 2023-08-15
Replace mockery to sinon library in mock-run which is used for Tasks Unit Tests

LiliaSabitova:
This PR includes changes that are made to the azure-pipelines-task-lib to replace the Mockery package with Sinon. Despite not utilizing the vulnerable method in Mockery (CVE-2022-37614 in Mockery 2.1.0), it was decided to make this change because Mockery is outdated. Sinon appeared to be the most suitable package. The changes proposed in this PR affect the unit tests of [Pipeline Tasks](https://github.com/microsoft/azure-pipelines-tasks), and some tasks may need adjustments when updating to the new version of task-lib. 

Changelog:

1. **Replaced Mockery with Sinon**: Sinon has been added as a new dependency, replacing Mockery in the project.
2. **Refactoring TaskMockRunner**: TaskMockRunner has been refactored to accommodate the changes brought by the transition from Mockery to Sinon. The new methods have been adjusted accordingly.

Impact of Changes:

The changes require additional alterations in several situations when updating [Pipeline Tasks](https://github.com/microsoft/azure-pipelines-tasks) to the new version of task-lib:

1. **Relative Path**: If a relative path was used for mocking a local module, changes are required.
2. **Import Aliasing**: If import aliasing was used in the task implementation, and it is being mocked in unit tests, changes are needed.
3. **Initialization Order**: Changes may occur because the sequence in which you initialize the task-lib and set inputs in tests is important. Setting inputs should go before initialization of task-lib.
4. **Module Clone and Recursive Calls**: To maintain the original behavior within a mocked function in Sinon, you need to add module's clone, otherwise it may lead to recursive calls.
5. **Localization Changes**: If a localized error message was used, it will now go through a mocked loc function and hence, get a mocked localization, not the real one.

Testing:

The changes were thoroughly tested using the unit tests of tasks under our ownership in the [Pipeline Tasks repository](https://github.com/microsoft/azure-pipelines-tasks). Out of around 200+ tasks, half are under our ownership. Out of these, not all are affected by these changes. Around 40-50 tasks that were affected have been tested.

Documentation:

After the PR gets merged, the documentation for the TaskMockRunner class should be updated to reflect the new and refactored methods and the usage of Sinon.js over Mockery. It should provide examples of how to use the new methods to accommodate the changes made.

Note:

This PR doesn't directly handle the localization changes, module initialization, relative path adjustments, recursive calls, and import aliasing. A separate migration guide will be released to help others navigate the necessary modifications in their tasks.

### DIFF
--- a/node/mock-run.ts
+++ b/node/mock-run.ts
@@ -71,7 +71,7 @@ export class TaskMockRunner {
 
         // Check if the method is a function
         if (typeof method !== 'function') {
-            console.warn(`WARNING: ${methodName} of ${newModule} is not a function. There is no option to replace getter/setter in this implementation. You can consider changing it.`);
+            console.log(`WARNING: ${methodName} of ${newModule} is not a function. There is no option to replace getter/setter in this implementation. You can consider changing it.`);
             return false;
         }
 


node/mock-run.ts
```diff
@@ -92,10 +136,21 @@ export class TaskMockRunner {
                     tlm[key] = this._exports[key];
                 });
 
-            mockery.registerMock('azure-pipelines-task-lib/task', tlm);
+
+            var tlt = require('azure-pipelines-task-lib/task');
+            for (let method in tlt) {
```
KonstantinTyukalov, line 141: nit: probably `const` might be better here

node/mock-run.ts
```diff
@@ -92,10 +136,21 @@ export class TaskMockRunner {
                     tlm[key] = this._exports[key];
                 });
 
-            mockery.registerMock('azure-pipelines-task-lib/task', tlm);
+
+            var tlt = require('azure-pipelines-task-lib/task');
```
KonstantinTyukalov, line 140: nit: const/let

node/mock-run.ts
```diff
@@ -44,9 +80,22 @@ export class TaskMockRunner {
     * @param val        Mock implementation of the module.
     * @returns          void
     */
-    public registerMock(modName: string, mod: any): void {
+    public registerMock(modName: any, mod: any): void {
```
KonstantinTyukalov, line 83:
```suggestion
    public registerMock(mod: string | object, modMock: object): void {
```

node/mock-run.ts
```diff
@@ -32,10 +35,43 @@ export class TaskMockRunner {
...
+
+    checkIsMockable(newModule, methodName, oldModule) {
+        const method = newModule[methodName];
+
+        if (!newModule.hasOwnProperty(methodName) || typeof method === 'undefined') {
+            return false;
+        }
+
+        if (typeof method !== 'function') {
+            console.log(`WARNING: ${methodName} of ${newModule} is not a function. There is no option to replace getter/setter in this implementation. You can consider changing it.`);
```
KonstantinTyukalov, line 64: probably, [console.warn](https://nodejs.org/api/console.html#consolewarndata-args) (but nor sure if we want to redirect to stderr)

node/package.json
```diff
@@ -32,6 +32,7 @@
     "q": "^1.5.1",
```
chris-tuncap, line 32: Might want to remove the mockery dependency to resolve the vulnerability.

node/mock-run.ts
LiliaSabitova: replaced by let

lgmorand:
:eyes: eagerly waiting for this to be merged <3 :)

jessehouwing:
Fixes: #955

PR #852, opened 2022-07-27
Added node16 to task.schema.json

AndreyIvanov42:
## Task description

Node16 must be specified as a valid property in `task.schema.json`.

## Changelog

- Added Node16 to `task.schema.json`
- Added MockTestRunner with Node16

PR #844, opened 2022-06-06
Migrating to Node 16

AndreyIvanov42:
## Task description

In order to migrate tasks to the Node16 execution handler, we have to update the pipelines-task-lib itself to support the Node16 runtime.

## Changelog

- Rewrote the pipeline to use node 16
- Updated dev dependencies
- Updated package-lock.json to version 2 to unlock the ability for deterministic and reproducible builds
- Fixed types for tests
- Fixed tool runner test to utilize changed output from Node 16
- Fixed `decipher.update` method call

## Testing

- Changes have been tested on Linux, MacOS, Windows: 
  by unit-tests
  manually as part of some task. Used tasks and cases are described [here](https://github.com/microsoft/azure-pipelines-task-lib/pull/844#issuecomment-1169891938)
- Enough test coverage for changes and current test coverage for the task doesn't look poor
- I didn't find any breaking changes
- There are no other concerns
- I haven't discovered any new uncovered test/use cases

Added unit tests: No

PR #962, opened 2023-08-18
Migrate to Node 20

qianz2:
- Added node20 to task.schema.json
- Update node 20 version in pipelines
- Update node 20 version in package.json and package-loc.json

Similar PRs for migrating to node16: 
https://github.com/microsoft/azure-pipelines-task-lib/pull/852
https://github.com/microsoft/azure-pipelines-task-lib/pull/844

### DIFF
--- a/node/docs/minagent.md
+++ b/node/docs/minagent.md
@@ -30,6 +30,8 @@ Use the details below to determine when specific agent features were added:
   - Added in 2.144.0. Used node v10.x
 * `node16` handler
   - Added in 2.206.1. Used node v16.x
+* `node20` handler
+  - Added in 2.225.0. Used node v20.x
 * `powershell3` handler
   - Added in 1.95.1
   - Updated in 1.97 to propagate `Data` property for endpoints
--- a/node/mock-test.ts
+++ b/node/mock-test.ts
@@ -152,8 +152,11 @@ export class MockTestRunner {
             case 16:
                 downloadVersion = 'v16.13.0';
                 break;
+            case 20:
+                downloadVersion = 'v20.3.1';
+                break;
             default:
-                throw new Error('Invalid node version, must be 6, 10, or 16 (received ' + version + ')');
+                throw new Error('Invalid node version, must be 6, 10, 16, or 20 (received ' + version + ')');
         }
 
         // Install node in home directory if it isn't already there.
@@ -185,7 +188,10 @@ export class MockTestRunner {
         );
         const keys = Object.keys(execution);
         for (let i = 0; i < keys.length; i++) {
-            if (keys[i].toLowerCase() == 'node16') {
+            if (keys[i].toLowerCase() == 'node20') {
+                // Prefer node 20 and return immediately.
+                return 20;
+            } else if (keys[i].toLowerCase() == 'node16') {
                 // Prefer node 16 and return immediately.
                 return 16;
             } else if (keys[i].toLowerCase() == 'node10') {
--- a/node/package.json
+++ b/node/package.json
@@ -39,7 +39,7 @@
     "@types/minimatch": "3.0.3",
     "@types/mocha": "^9.1.1",
     "@types/mockery": "^1.4.29",
-    "@types/node": "^16.11.39",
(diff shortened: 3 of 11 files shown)


node/docs/minagent.md
```diff
@@ -30,6 +30,8 @@ Use the details below to determine when specific agent features were added:
   - Added in 2.144.0. Used node v10.x
 * `node16` handler
   - Added in 2.206.1. Used node v16.x
+* `node20` handler
+  - Added in 3.225.0. Used node v20.x
```
qianz2, line 34: We supported node20 handler in this [PR](https://github.com/microsoft/azure-pipelines-agent/pull/4346), and it's published in release [3.224.0](https://github.com/microsoft/azure-pipelines-agent/releases), so my understanding is that here we will require minimum agent version to be >= 3.224.0 to enable node20 support.

node/mock-test.ts
qianz2, line 1: Should we make node20 the default node version in this file? The default is node 16. https://github.com/microsoft/azure-pipelines-task-lib/blob/49efa4ed0a6239c361ecfc0209d84eb2806cd6ef/node/mock-test.ts#L177

node/package.json
```diff
@@ -39,7 +39,7 @@
     "@types/minimatch": "3.0.3",
     "@types/mocha": "^9.1.1",
     "@types/mockery": "^1.4.29",
-    "@types/node": "^16.11.39",
+    "@types/node": "^20.3.1",
     "@types/q": "^1.5.4",
     "@types/semver": "^7.3.4",
     "@types/shelljs": "^0.8.8",
```
qianz2, line 45: How to ensure that the devDependencies in package.json are bumped up as needed? Node20 required typescript version upgrade, we update typescript version to latest 5.1.6 in node 20 tasks, so I put same version here and then run 'npm update' command.

max-zaytsev:
@qianz2 how were these changes tested?

qianz2:
> @qianz2 how were these changes tested?

Hi @max-zaytsev, I didn't see ways to test these changes other than the existing check azure-pipelines-task-lib. I think we'll use this new release preview version in tasks, then running task tests to ensure they'll work with node20 tasks.

node/mock-test.ts
DmitriiBobreshev: Yes, we should since 5.x version uses for node20, if users needs only node16 support they can use 4.x version

node/package.json
DmitriiBobreshev: you could use some third-party tools like [npm-check](https://www.npmjs.com/package/npm-check) to validate that everything is fine

MirKml:
important notice - node20 isn't supported on Ubuntu 18.04 - see [nodesource](https://github.com/nodesource/distributions#debian-and-ubuntu-based-distributions)
But Ubuntu 18.04 is as supported [distro for self hosted agents](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/linux-agent?view=azure-devops#check-prerequisites)

When you setup node20 as default runner, it crashes all current production environments with Ubuntu 18.04.
It can be a huge problem.
Node16 runner must be default one for current azure pipeline agent.

qianz2:
The changes in this PR is tentative, all migrated tasks are working well without bumping and upgrading this task-lib, same as for the tool-lib [PR](https://github.com/microsoft/azure-pipelines-tool-lib/pull/207) that I initiated, we've planned to create release branches and check changes to master, but they're not blocking the migration as far as I know. So I will hold it for now, eventually we'll check it in.

tasks.schema.json
```diff
@@ -429,6 +435,9 @@
       "additionalProperties": false,
       "description": "Execution options for this task (on Post-Job stage)",
       "properties": {
+        "Node20": {
+          "$ref": "#/definitions/executionObject"
+        },
         "Node16": {
           "$ref": "#/definitions/executionObject"
```
merlynomsft, line 442: [not related to this line] releases/5.x is used for other changes - we'll need to re-target these changes

Drarig29:
Hi @merlynomsft @qianz2, is it possible to have a status update for this PR?

PR #979, opened 2023-10-10
Introduce tl.getNodeMajorVersion

onetocny:
This PR adds new function `getNodeMajorVersion` to tasks.ts.

**To do:**
- [x] Bump up package version
- [x] Update changelog

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -707,9 +707,9 @@ export function getPlatform(): Platform {
  * @returns {Number} Node's major version.
  */
 export function getNodeMajorVersion(): Number {
-    const version = process.versions.node;
+    const version = process?.versions?.node;
     if (!version) {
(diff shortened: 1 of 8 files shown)


node/task.ts
```diff
@@ -702,6 +702,24 @@ export function getPlatform(): Platform {
     }
 }
 
+/**
+ * Resolves major version of Node.js engine used by the agent.
+ * @returns {Number} Node's major version.
+ */
+export function getNodeMajorVersion(): Number {
+    const version = process.versions.node;
+    if (!version) {
```
KonstantinTyukalov, line 711: When can it be undefined? As I know it's possible in browser/other js runtimes.
Should we just throw in this case?

qianz2 (approved):

PR #1009, opened 2024-01-10
Add support for the optional 'source' property to the error/warning functions

DenisNikulin5:
Added support for the optional 'source' property to the error/warning functions to distinguish internal task issues and the issues that occur because of incorrect customer actions (for example, the invalid user script).

Testing: local dev environment with additional changes in PowerShell, Bash and CmdLine tasks.

Changed the version of npm packages:
node 4.7.0 -> 4.8.0
powershell 0.17.0 -> 0.18.0

### DIFF
--- a/node/internal.ts
+++ b/node/internal.ts
@@ -20,6 +20,13 @@ import crypto = require('crypto');
 export var _knownVariableMap: { [key: string]: _KnownVariableInfo; } = {};
 
 export var _vault: vm.Vault;
+//-----------------------------------------------------
+// Enums
+//-----------------------------------------------------
+export enum IssueSource {
+    CustomerScript = 'CustomerScript',
+    TaskInternal = 'TaskInternal'
+}
 
 //-----------------------------------------------------
 // Validation Checks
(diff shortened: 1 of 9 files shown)


vmapetr (approved):

node/internal.ts
```diff
@@ -282,12 +282,12 @@ export function _command(command: string, properties: any, message: string) {
...
 
-export function _warning(message: string): void {
-    _command('task.issue', { 'type': 'warning' }, message);
+export function _warning(message: string, source?: string): void {
+    _command('task.issue', { 'type': 'warning', 'source': source }, message);
 }
 
-export function _error(message: string): void {
-    _command('task.issue', { 'type': 'error' }, message);
+export function _error(message: string, source?: string): void {
```
KonstantinTyukalov, line 289: What sources are we expected to see?
Probably it's better to use enum/template type

DenisNikulin5: Expected sources are something like 'customerscript', 'taskinternal'

DenisNikulin5: Okay, probably you're right. I'll re-implement this thing, no big deal

KonstantinTyukalov (approved):

DergachevE (approved):

PR #1010, opened 2024-01-16
Added feature helpers

KonstantinTyukalov:
Added wrappers for tasks feature flags.
PowerShell: `Get-PipelineFeature`
Node: `getPipelineFeature`

WI: [2041738](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2041738/)


node/task.ts
```diff
@@ -801,7 +824,7 @@ export function mkdirP(p: string): void {
     let testDir: string = p;
     while (true) {
         // validate the loop is not out of control
-        if (stack.length >= (process.env['TASKLIB_TEST_MKDIRP_FAILSAFE'] || 1000)) {
+        if (stack.length >= Number(process.env['TASKLIB_TEST_MKDIRP_FAILSAFE'] || 1000)) {
```
DergachevE, line 827: What is `1000` value for? How was it defined?

DergachevE (approved):

node/task.ts
KonstantinTyukalov: To be clear, the whole method logic is a bit confusing for me at the moment😄 
I'm not sure why we manually create directories by stack rather than just call the `fs.mkdirSync(p, {recursive: true} );`
Probably, there is some historical reasons for it, but we can consider removing this method in later PRs.

PR #1021, opened 2024-02-15
Add task SDK token support

DenisNikulin5:
This PR added support for task SDK token. It adds the ability to distinguish between calls to the task.issue command using the task-lib and manual calls caused by user scripts.

Also:
- Added unit tests from the changes
- Fixed stout handler for PsRunner
- Bumped the package versions:
PowerShell: 0.19 -> 0.20
Node: 4.9.1 -> 4.10.0

Related PR: https://github.com/microsoft/azure-pipelines-agent/pull/4662

### DIFF
--- a/node/internal.ts
+++ b/node/internal.ts
@@ -51,6 +51,20 @@ export function _endsWith(str: string, end: string): boolean {
     return str.slice(-end.length) == end;
 }
 
+export function _truncateBeforeSensitiveKeyword(str: string, sensitiveKeywordsPattern: RegExp): string {
+    if(!str) {
+        return str;
+    }
+
+    const index = str.search(sensitiveKeywordsPattern); 
+
+    if (index <= 0) {
+        return str;
+    }
+
+    return `${str.substring(0, index)}...`;
+}
+
 //-----------------------------------------------------
 // General Helpers
 //-----------------------------------------------------
--- a/node/internal.ts
+++ b/node/internal.ts
@@ -20,13 +20,7 @@ import crypto = require('crypto');
 export var _knownVariableMap: { [key: string]: _KnownVariableInfo; } = {};
 
 export var _vault: vm.Vault;
-
-let _taskSdkToken: string;
-if (process.env["TASK_SDK_COMMAND_TOKEN"])
-{
-    _taskSdkToken = String(process.env["TASK_SDK_COMMAND_TOKEN"]);
-    delete process.env["TASK_SDK_COMMAND_TOKEN"];
-}
+var _taskSdkToken: string;
 
 //-----------------------------------------------------
 // Enums
@@ -747,6 +741,10 @@ export function _loadData(): void {
         }
     }
     _debug('loaded ' + loaded);
+
+    let token = process.env["TASK_SDK_COMMAND_TOKEN"];
+    delete process.env["TASK_SDK_COMMAND_TOKEN"];
+    _taskSdkToken = token ? String(token) : "";
 
     // store public variable metadata
     let names: string[];
--- a/node/internal.ts
+++ b/node/internal.ts
@@ -21,9 +21,10 @@ export var _knownVariableMap: { [key: string]: _KnownVariableInfo; } = {};
 
 export var _vault: vm.Vault;
 
-let _taskSdkToken: string = String(process.env["TASK_SDK_COMMAND_TOKEN"]);
-if (_taskSdkToken)
+let _taskSdkToken: string;
+if (process.env["TASK_SDK_COMMAND_TOKEN"])
 {
+    _taskSdkToken = String(process.env["TASK_SDK_COMMAND_TOKEN"]);
     delete process.env["TASK_SDK_COMMAND_TOKEN"];
 }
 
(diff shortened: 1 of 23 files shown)


node/test/taskissuecommandtests.ts
```diff
@@ -0,0 +1,106 @@
...
+        process.env['AGENT_VERSION'] = '2.115.0'
+        let variable = tl.getVariable('TASK_SDK_COMMAND_TOKEN');
+        let taskVariable = tl.getTaskVariable('TASK_SDK_COMMAND_TOKEN');
+        assert.equal(variable, undefined);
+        assert.equal(taskVariable, undefined);
+        
+        done();
+    })
+
+    it('adds the token for task.issue messages', function (done) {
```
KonstantinTyukalov, line 50: what if token wasn't found in env?

DenisNikulin5: If the token wasn't found, it would be absent in the command property list.

Just like this:
https://github.com/microsoft/azure-pipelines-task-lib/blob/f1481180c4673ed739197a571ce9544bd3cc8d47/node/test/resulttests.ts#L50

KonstantinTyukalov: could you please cover this case by test as well (and for powershell)

DenisNikulin5: Okay.

That's already implemented for PowerShell with and without a token using
```
powershell/Tests/L0/Write-LoggingCommand.WritesTaskIssue.ps1
powershell/Tests/L0/Write-LoggingCommand.WritesTaskIssueWithToken.ps1
```

DenisNikulin5: Done

KonstantinTyukalov (approved):

DmitriiBobreshev (approved):

DergachevE (approved):

powershell/VstsTaskSdk/VstsTaskSdk.psd1
```diff
@@ -9,7 +9,7 @@
     PowerShellVersion = '3.0'
     FunctionsToExport = '*'
     CmdletsToExport = ''
-    VariablesToExport = 'IssueSources'
+    VariablesToExport = '*'
```
merlynomsft, line 12: Not sure the implication of changing this to "*" -- would it be safer to make more narrow?

powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
```diff
@@ -554,7 +560,7 @@ function Write-LogIssue {
         [switch]$AsOutput,
         [AllowNull()]
         [ValidateSet('CustomerScript', 'TaskInternal')]
-        [string]$IssueSource)
+        [string]$IssueSource = $IssueSources.TaskInternal)
```
merlynomsft, line 563: What if a non-builtin task task calls Write-LogIssue.  Will "IssueSource" be ignored?

node/internal.ts
```diff
@@ -303,12 +305,12 @@ export function _command(command: string, properties: any, message: string) {
...
 
-export function _warning(message: string, source?: IssueSource): void {
-    _command('task.issue', { 'type': 'warning', 'source': source }, message);
+export function _warning(message: string, source: IssueSource = IssueSource.TaskInternal): void {
+    _command('task.issue', { 'type': 'warning', 'source': source, 'token': _taskSdkToken }, message);
 }
 
-export function _error(message: string, source?: IssueSource): void {
-    _command('task.issue', { 'type': 'error', 'source': source }, message);
+export function _error(message: string, source: IssueSource = IssueSource.TaskInternal): void {
```
merlynomsft, line 312: what if a non-built in task calls _error.  is IssueSource ignored?

I'm thinking a better default for this would be "Unspecified" rather than TaskInternal or CustomerSource.  Then we can deal with the default behavior upstream.  Otherwise, changing the default behavior will require more churn.

I anticipate we'll want to change the default later...  I understand choosing "TaskInternal" for now to reduce churn.  However, we may want to change the default to opt-in.

powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
DenisNikulin5: It won't be ignored, a user can see this property if the pipeline is executing with system diagnostic enabled (debug mode). However, it won't affect telemetry since ADO ignores issues for 3rd party tasks at all.

node/internal.ts
DenisNikulin5: I think we can keep it as it is in the task-lib repo. It's easy to implement the behavior you've described on the agent side. Because the agent knows if the task is in-the-box or not but the task itself doesn't.

powershell/VstsTaskSdk/VstsTaskSdk.psd1
DenisNikulin5: Even though it is specified here that we export all members it isn't so because we explicitly export members in the psm1 file. It's completely safe. The same works for function export.
https://github.com/microsoft/azure-pipelines-task-lib/blob/4ee6f874499c3b5cfc599871f91cb52b3a3fcd23/powershell/VstsTaskSdk/VstsTaskSdk.psm1#L98-L100

PR #1022, opened 2024-02-15
Add setSanitizedResult function

onetocny:
There were several cases when we observed sensitive data (such as passwords) leakage through `task.setResult` function. To prevent such behavior this PR introduces `setSanitizedResult` function that truncates the passed message before first occurrence of well known sensitive keyword. For more details see the original [incident](https://portal.microsofticm.com/imp/v3/incidents/incident/466877337/summary) or its [repair item](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2147925).

### DIFF
--- a/node/internal.ts
+++ b/node/internal.ts
@@ -49,6 +49,20 @@ export function _endsWith(str: string, end: string): boolean {
     return str.slice(-end.length) == end;
 }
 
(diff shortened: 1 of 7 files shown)


DenisNikulin5 (approved):
LGTM 👍

PR #1030, opened 2024-04-02
Fix toolrunner line buffering

EvanCahill:
The `_processLineBuffer` needs to buffer data between method calls but the variable it is using as its buffer is a string parameter and thus goes out of scope at the end of the method. To ensure the buffer is persisted across calls it needs to be an object.


KonstantinTyukalov:
Reopened here: #1055

PR #1050, opened 2024-07-12
Use tar tool for .tar.gz node.js archives

KonstantinTyukalov:
On Linux and MacOs, we download `.tar.gz` archives rather than `.zip` for Windows. `adm-zip` package does not support the `.tar.gz` and throws the error:
```text
Invalid or unsupported zip format. No END header found
```

Here I used the pre-installed `tar` tool to unpack node for unix systems

WI: [AB#2190598](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2190598)

### DIFF
--- a/node/buildutils.js
+++ b/node/buildutils.js
@@ -51,7 +51,7 @@ const getExternalsAsync = async () => {
                 downloadFileAsync(nodeUrl + '/' + nodeVersion + '/win-x64/node.exe'),
                 downloadFileAsync(nodeUrl + '/' + nodeVersion + '/win-x64/node.lib')
             ]);
-            
+
             var nodeDirectory = path.join(testPath, 'node');
             mkdir('-p', nodeDirectory);
             cp(nodeExePath, path.join(nodeDirectory, 'node.exe'));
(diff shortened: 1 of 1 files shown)


node/buildutils.js
```diff
@@ -130,8 +130,16 @@ var downloadArchiveAsync = async function (url, fileName) {
...
     mkdir('-p', targetPath);
-    var zip = new admZip(archivePath);
-    zip.extractAllTo(targetPath);
+
+    if (targetPath.endsWith('.zip')) {
+        var zip = new admZip(archivePath);
+        zip.extractAllTo(targetPath);
+    }
+    else if (targetPath.endsWith('.tar.gz')) {
+        run(`tar --extract --gzip --file="${archivePath}" --directory="${targetPath}"`);
```
ismayilov-ismayil, line 139: Should we check that 'tar' exist on the OS?

KonstantinTyukalov: I believe if it's not, it will fail with verbose message 😃

vmapetr (approved):

node/buildutils.js
vmapetr: AFAIK, tar comes preinstalled on all distros that we currently support. Same goes for MacOS.

kirill-ivlev (approved):

DergachevE (approved):

PR #1056, opened 2024-08-13
Removing debug log while Setting Resource path

EshitaD:


node/internal.ts
```diff
@@ -195,7 +195,7 @@ export function _setResourcePath(path: string, ignoreWarnings: boolean = false):
     }
     else {
         if (ignoreWarnings) {
-            _debug(_loc('LIB_ResourceFileAlreadySet', path));
+           // _debug(_loc('LIB_ResourceFileAlreadySet', path));
```
z00sts, line 198: please delete the line, no need to keep commented lines

z00sts (changes requested):

node/internal.ts
EshitaD: Hi @z00sts, I have removed the line.

z00sts (approved):

PR #1080, opened 2024-11-28
Fix process invocation for PowerShell handler

DenisNikulin5:
Recently we faced [the issue](https://github.com/microsoft/azure-pipelines-tasks/issues/19781) where exit code for process was not set, and it was the root cause of pipeline failures.

I noticed that PowerShell could behave incorrectly if we manage the process using .NET API directly (for instance, we had to cache the process handle to obtain the status code as a workaround. please check https://stackoverflow.com/a/23797762)

Even though I didn't manage to reproduce the issue specified at the beginning by myself. I suppose that native PS cmdlets will help to fix it. 

WI: [AB#2222447](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2222447)

Testing: manual, unit tests

Other changes: Changed ts version to fix the unit tests

### DIFF
--- a/powershell/VstsTaskSdk/ToolFunctions.ps1
+++ b/powershell/VstsTaskSdk/ToolFunctions.ps1
@@ -202,13 +202,8 @@ function Invoke-Process {
             $processOptions.Add("RedirectStandardError", $StdErrPath)
         }
 
-        # TODO: For some reason, -Wait is not working on agent.
-        # Agent starts executing the System usage metrics and hangs the step forever.
-        $proc = Start-Process @processOptions
-
-        # https://stackoverflow.com/a/23797762
-        $null = $($proc.Handle)
-        $proc.WaitForExit()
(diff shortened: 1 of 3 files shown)


KonstantinTyukalov (approved):

vmapetr (approved):

kirill-ivlev (approved):

tarunramsinghani (approved):

chrisfcarroll:
May I ask, is there a way to see other than try it out in a pipeline whether this merge is live on dev.azure.com ? 

The MSBuild and VSBuild tasks in which I experienced this error don't have a visible version bump. Or is that irrelevant?

tarunramsinghani:
Seems this version is not updated the tasks yet so the MSBuild task does not have this fix for now. They need to be updated and will add them to be part of next release of the tasks.

PR #1105, opened 2025-05-28
Replace Array.flat() with custom flattenArray() function for Node.js compatibility

rishabhmalikMS:
**Problem**
The current code uses Array.flat() which is only available in Node.js versions >= 12. This causes compatibility issues when running on older Node.js versions that are still supported by Azure Pipelines agents.

**Changes**

- Added new flattenArray() helper function to recursively flatten nested arrays
- Removed usage of Array.flat() method

**Github Issue Link:** https://github.com/microsoft/azure-pipelines-tasks/issues/21048

**Testing**
Tested with already existing unit tests having multiple paths as input to LS function in an array.

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -1047,7 +1047,7 @@ export function ls(optionsOrPaths?: unknown, ...paths: unknown[]): string[] {
 
     // Flatten paths if the paths argument is array
     if (Array.isArray(paths)) {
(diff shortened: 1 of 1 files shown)


sanjuyadav24:
changes looks good, can we please update canary test pipeline to have this test case

PR #1106, opened 2025-05-29
LS function fix for single file input

rishabhmalikMS:
Explanation of the fix we implemented in the `ls` function for the single file input issue:

### Issue
- When providing a single file as input to `ls()`, it was returning an empty string because the relative path calculation was comparing the file path against itself.

### Fix Details
- We separated the handling of files and directories into two phases:
  ```typescript
  // Phase 1: Early file handling
  if (stats.isFile()) {
      // For files, just add the basename to results array
      const fileName = path.basename(pathEntry);
      results.push(fileName);
  } else {
      // Keep directories for later processing
      remainingPaths.push(pathEntry);
  }
  ```

### Key Changes
1. Files are processed first, before any directory traversal
2. For files, we simply extract and store the basename
3. Only directories continue to the existing relative path calculation logic

### Benefits
1. No changes to existing directory handling code
2. Maintains backward compatibility
3. Fixes the single file issue in ArchiveFiles@2 task
4. Clearer separation between file and directory processing

### Example
```yaml
- task: ArchiveFiles@2
  inputs:
    rootFolderOrFile: 'path/to/single/file.txt'
```
Now correctly adds the file to the archive with its proper name instead of an empty string.

### Link to modified canary pipeline with specific file archiving tests. 
This pipeline run is from my private organization which is run on archive file task linked locally to modified task-lib code. This includes new tests named `Test archive with specific file zip` for testing the change
https://dev.azure.com/rishabhmalikOrg/Test%20Project/_build/results?buildId=1040&view=logs&j=1809f88b-43e1-5a25-3102-9dca9646ed63

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "5.2.0",
+  "version": "5.2.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "5.2.0",
+  "version": "5.2.1",
   "description": "Azure Pipelines Task SDK",
   "main": "./task.js",
   "typings": "./task.d.ts",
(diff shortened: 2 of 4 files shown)


raujaiswal:
After running npm install with Node.js version v20.19.1, there are changes detected in the package-lock.json file. Could you please review these changes and confirm whether they are necessary to commit?

rishabhmalikMS:
> After running npm install with Node.js version v20.19.1, there are changes detected in the package-lock.json file. Could you please review these changes and confirm whether they are necessary to commit?
> 
> image

These changes are mostly due to dependencies version being specified with `^`

PR #1079, opened 2024-11-14
Remove the shelljs dependency

ivanduplenskikh:
**Changes**: This PR changes the shelljs methods to native NodeJS methods

**Related WI**: [AB#2237714](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2237714)

PR #1146, opened 2025-12-23
Fix ExtractFile Node 24 migration issue

rajmishra1997:
Issue:
Canary Test pipeline failed after Node 24 migartion for Task ExtractFile: https://dev.azure.com/canarytest/PipelineTasks/_build/results?buildId=235044&view=results

[AB#2343422](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2343422)
AB#[2343402](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2343402)

Why this issue:

- As part of node 24 migration, task-lib version is changed from 4.16.0 to 5.2.2
- Issue arose from line [1123](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/task.ts#L1123)
- task.ts ls functionality changed in PR: https://github.com/microsoft/azure-pipelines-task-lib/pull/1079, Earlier version used shell.js
- Line 1123, pathsCopy.find, which is Array.Prototype.Find should return undefined in case of no match but returned -1 
- This issue was caused by a custom polyfill implementation overriding the native behavior, as shown in the image below.
arrayfindIssue

Solution:
1. Added method "safeFind" in place of Array.Prototype.Find
2. Tested in DevFabric

Failure Scenario:
extractFileIssue

Successful Scenario
ExtractFileFixed

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "5.2.4",
+  "version": "5.2.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "5.2.4",
+  "version": "5.2.5",
   "description": "Azure Pipelines Task SDK",
   "main": "./task.js",
   "typings": "./task.d.ts",


node/task.ts
```diff
@@ -1069,7 +1069,7 @@ export function ls(optionsOrPaths?: unknown, ...paths: unknown[]): string[] {
     if (paths.length === 0) {
         paths.push(path.resolve('.'));
     }
-    const pathsCopy = [...paths];
+    const pathsCopy = [...paths] as string[];
```
sanjuyadav24: shouldn't we fix this in the function signature of ls where path is defined as unknow?

Issue #1166, opened 2026-04-09
`cp()` treats `@` and `#` in file paths as glob patterns, causing infinite recursion (RangeError: Maximum call stack size exceeded)


Ludde321:
## Summary

The `cp` function in `node/task.ts` uses a regex to detect glob patterns in the source path:

```typescript
const hasGlobPattern = /[*?{[!@#]/.test(source);
```

**The `@` and `#` characters are not glob pattern characters**, but this regex treats them as such. When a file path contains `@` (very common on Linux — e.g., a service account named `user@domain.com`), the `cp` function incorrectly enters the glob-handling branch, which calls `findMatch()` and then **recursively calls `cp()` with the same resolved path**, causing infinite recursion:

```
RangeError: Maximum call stack size exceeded
```

### Root Cause

The call chain is:

1. `cp(source, dest, '-r')` — `source` contains `@`
2. `hasGlobPattern` regex matches `@` → enters glob branch
3. `findMatch(sourceDir, [basename])` resolves the path back to the **same literal file** (because `@` is not actually a glob)
4. `cp(resolvedPath, dest, ...)` is called recursively
5. The resolved path still contains `@` → goto step 2
6. **Infinite recursion** → `RangeError: Maximum call stack size exceeded`

The relevant code ([`node/task.ts`](https://github.com/microsoft/azure-pipelines-task-lib/blob/5d545c0eb2882e87022007b9acd2f386b9daee21/node/task.ts)):

```typescript
const hasGlobPattern = /[*?{[!@#]/.test(source);
if (hasGlobPattern) {
    let sourcesToProcess: string[] = [];
    let sourceDir = path.dirname(source);
    sourceDir = sourceDir == '.' ? path.resolve() : sourceDir;
    sourcesToProcess = findMatch(sourceDir, [path.basename(source)]);
    // ...
    for (const src of sourcesToProcess) {
        cp(src, destination, options as CopyOptionsVariants, continueOnError, retryCount);
        // ↑ recursive call with the same path still containing @
    }
    return;
}
```

Only `*`, `?`, `{`, `[`, and `!` are meaningful glob characters in [minimatch](https://github.com/isaacs/minimatch) (which this codebase uses). The `@` and `#` should be removed from the regex.

### Introduced By

[...]

PR #1167, opened 2026-04-14
Fix glob patterns in the cp function

Deekshitha2812:
### Problem
 The `cp` function's glob pattern regex (`/[*?{[!@#]/.test(source)`) incorrectly treated a literal `@` character in file paths as an extglob pattern. The `@` symbol is only a glob operator when used as `@(...)`, but the old regex matched a standalone `@` anywhere in the path. This caused the function to enter the glob resolution branch, leading to an infinite loop and `RangeError: Maximum call stack size exceeded`.
 
 ### Fix
 Updated the glob detection regex to correctly distinguish between:
 - **Glob operators**: `*`, `?`, `{`, `[` — matched as before
 - **`#` comments**: only at the start of the string (`/^[#!]/.test(source)`)
 - **Extglob `@(...)`**: requires `@` (or `+`,`!`) followed by `(` (`/[@+!]\(/.test(source)`)
 
 A standalone `@` in a path (e.g., `dir@test/file.txt`) is no longer 
 misidentified as a glob pattern.
 
 ### Changes
 - **`node/task.ts`** — Updated `hasGlobPattern` regex in `cp` function
 - **`node/package.json`** — Version bump 5.2.9 → 5.2.10
 - **`node/tests/cp.test.ts`** — Added tests for:
   - `@()` extglob pattern matching (e.g., `@(file1|file2).txt`)
   - Literal `@` in directory paths (e.g., `dir@test/file.txt`)
 
 ### Fixes
 - #1166
 - microsoft/azure-pipelines-tasks#21987

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -1232,7 +1232,7 @@ export function cp(sourceOrOptions: unknown, destinationOrSource: string, option
         if (!fs.existsSync(destination) && !force) {
             throw new Error(loc('LIB_PathNotFound', 'cp', destination));
         }
-        const hasGlobPattern =  /[*?{[!@#]/.test(source);
+        const hasGlobPattern =  /[*?{!\[]/.test(source) || /^[#]/.test(source) || /[@+]\(/.test(source);
         if (hasGlobPattern) {
             let sourcesToProcess: string[] = [];
             let sourceDir = path.dirname(source);
--- a/node/task.ts
+++ b/node/task.ts
@@ -1232,14 +1232,14 @@ export function cp(sourceOrOptions: unknown, destinationOrSource: string, option
         if (!fs.existsSync(destination) && !force) {
             throw new Error(loc('LIB_PathNotFound', 'cp', destination));
         }
-        const hasGlobPattern =  /[*?{!\[]/.test(source) || /^[#]/.test(source) || /[@+]\(/.test(source);
-        if (hasGlobPattern) {
+        const isPattern = /[*?{\[]/.test(source) || /^[#!]/.test(source) || /[@+!]\(/.test(source);
+        if (isPattern) {
             let sourcesToProcess: string[] = [];
             let sourceDir = path.dirname(source);
             sourceDir = sourceDir == '.' ? path.resolve() : sourceDir;
             sourcesToProcess = findMatch(sourceDir, [path.basename(source)]);
             if (sourcesToProcess.length === 0) {
-                debug(`No matches found for glob pattern: ${source}`);
+                debug(`No matches found for pattern: ${source}`);
             }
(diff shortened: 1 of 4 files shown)


sanjuyadav24 (approved):

node/task.ts
```diff
@@ -1232,7 +1232,7 @@ export function cp(sourceOrOptions: unknown, destinationOrSource: string, option
         if (!fs.existsSync(destination) && !force) {
             throw new Error(loc('LIB_PathNotFound', 'cp', destination));
         }
-        const hasGlobPattern =  /[*?{[!@#]/.test(source);
+        const hasGlobPattern =  /[*?{!\[]/.test(source) || /^[#]/.test(source) || /[@+]\(/.test(source);
```
tarunramsinghani, line 1235: Following suggestions from copilot, please have a look ince 

-  1. Missing extglobs ?( and !() — It catches @( and +( but misses ?(...), !(...) as extglob (note: bare ! and ? are caught by the first regex, but only coincidentally — ! outside [...] 
  isn't always a glob, and ?( specifically is the extglob form).
- Bare ! false positive — ! is only meaningful in globs inside brackets like [!abc] (negation). The pattern matches ! anywhere, so a string like "hello! world" would be flagged as a 
  glob pattern.
- Bare ? in non-glob contexts — ? in URLs or query strings (file?param=1) would trigger a false positive.
- Bare { without } — {a,b} is a brace expansion, but a lone { in text like "function() {" would be a false positive.
-

Deekshitha2812: Bare ? and { are valid glob operators (file?.txt, {a,b}.txt), but standalone occurrences in non-glob contexts (e.g., file?param=1, function() {) would be false positives. However, since cp operates on file paths — not URLs or code strings — these cases are unlikely to occur in practice. Removing them from the regex would risk missing legitimate glob patterns, so they remain.

  Bare ! is issue. We can update the regex to `  /[*?{\[]/.test(source) || /^[#!]/.test(source) || /[@+!]\(/.test(source)`. ! can appear in several contexts:

   - !(pattern) — extglob negation, valid glob → detected by /[@+!]\(/.test(source)
   - [!abc] — character class negation, valid glob → already detected by [ in /[*?{\[]/.test(source)
   - !file.txt — negation/exclude at the start → detected by /^[#!]/.test(source)
   - hello! world — literal ! in a filename, not a glob → with ! removed from the first character class regex, this is no longer a false positive
 
  The !file.txt case in cp will always produce empty results. A leading ! means "exclude from existing results," but cp only accepts a single pattern — there are no prior include results to exclude from.

tarunramsinghani (approved):

rishabhmalikMS (approved):

PR #1168, opened 2026-04-18
Fix console color fallback handling for invalid host values

MJECloud:
# Description
 Improve PowerShell logging color handling by adding a safe console color fallback path. This change validates host-provided color values and falls back to known defaults when values are invalid, preventing runtime failures when writing task issue and debug output.

# Why? 
Some Linux environments can return invalid console color values (for example -1), which caused task execution to fail when Write-Host received an unsupported color. This update makes logging resilient across platforms and avoids breaking pipelines due to host color metadata.

# Related
https://github.com/microsoft/azure-pipelines-agent/pull/5530

### DIFF
--- a/powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
+++ b/powershell/VstsTaskSdk/LoggingCommandFunctions.ps1
@@ -19,6 +19,22 @@ $IssueSources = @{
     TaskInternal = "TaskInternal"
 }
 
+function Get-ConsoleColorOrDefault {
+    [CmdletBinding()]
+    param(
+        [AllowNull()]
+        $Value,
+        [Parameter(Mandatory = $true)]
+        [System.ConsoleColor]$DefaultColor
+    )
+
+    if ($Value -is [System.ConsoleColor] -and [System.Enum]::IsDefined([System.ConsoleColor], $Value)) {
+        return $Value
+    }
+
+    return $DefaultColor
+}
+
 <#
 .SYNOPSIS
 See https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md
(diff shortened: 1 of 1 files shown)


tristanbarcelon:
@MJECloud, when I wrote Windows Powershell custom tasks, I ended up using the [powershell task sdk](https://github.com/microsoft/azure-pipelines-task-lib/tree/master/powershell), though it is only for Windows. @tarunramsinghani, @geekzter do we need a separate vststasksdk for pwsh handler once it is merged so pwsh native tasks will execute on windows and linux?

MJECloud:
@tristanbarcelon i tested all my changes on windows and ubutnu. It worked, so I think the module isn't windows only, since the binaries are built on .netstandard2.1

tarunramsinghani:
The Powershell module currently is specifically targets Windows and Linux or Macos are not the tagets for this mdoule. I was looking around last week for small time and for making new pwsh task handler and this module compatible cross platform, we will need to do significant investment in identifying and fixing issues, one being what is mentioned above for targetting ".netstandard 2.1".

MJECloud:
@tarunramsinghani I'm sure there are still some investment. From what i saw is that the module doesnt target windows only. I can import it and use it on my ubuntu with PowerShell 7.6

tarunramsinghani:
I am sure it can be imported but not all fucntionality will work as there is code like this that specifically will only work on windows see for e.g. [kernel32.dll reference](https://github.com/microsoft/azure-pipelines-task-lib/blob/70fc78360b31db1e0dabdc4a9611564b88f3f13a/powershell/CompiledHelpers/VstsTaskSdk.cs#L111) and there are other specific instances where it will break on other platforms. 

I am moving this change with version bump to a different PR as this looks safe for merging. #1186

MJECloud:
@tarunramsinghani Ah i missed that part, thank you! 😊

PR #1176, opened 2026-06-03
Fix infinite loop when using `[` or `]` in paths

### LINKED ISSUE #1166: `cp()` treats `@` and `#` in file paths as glob patterns, causing infinite recursion (RangeError: Maximum call stack size exceeded)

nielswitte:
## Problem
Using `[` and/or `]` in a path (directory or filename) will result in an endless loop causing: Maximum call stack size exceeded

## Fix
Stop the loop when the matched value is the same as the source value. 

## Changes
- `node/task.ts` — Additional `if` statement added  to check if the new `src` in `sourcesToProcess` differs from the original `source`
- `node/tests/cp.ts` — Test added with `[` and `]` in directory and filename

## Fixes
#1166

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -1242,6 +1242,15 @@ export function cp(sourceOrOptions: unknown, destinationOrSource: string, option
                 debug(`No matches found for pattern: ${source}`);
             }
             for (const src of sourcesToProcess) {
+                if (path.resolve(src) === path.resolve(source)) {
+                    // Literal path that was mistakenly treated as a glob pattern.
+                    // Copy directly without recursing to prevent infinite recursion.
+                    const dest = fs.existsSync(destination) && fs.lstatSync(destination).isDirectory()
+                        ? path.join(destination, path.basename(src))
+                        : destination;
+                    copyWithPreservedSymlinks(src, dest, force);
+                    continue;
+                }
                 cp(src, destination, options as CopyOptionsVariants, continueOnError, retryCount);
             }
             return;
--- a/node/test/cp.ts
+++ b/node/test/cp.ts
@@ -349,4 +349,20 @@ describe('cp cases', () => {
     tl.rmRF('dirD');
     done();
   });
+
+  it('cp handles literal [ and ] in path without infinite recursion', (done) => {
+    const srcDir = path.resolve(DIRNAME, '[Test] dir');
+    const destDir = path.resolve(DIRNAME, '[d]est');
+    tl.mkdirP(srcDir);
+    tl.mkdirP(destDir);
+    fs.writeFileSync(path.join(srcDir, 'file.txt'), 'content');
+
+    assert.doesNotThrow(() => tl.cp(path.join(srcDir, 'file.txt'), destDir));
+    assert.ok(fs.existsSync(path.join(destDir, 'file.txt')));
+    assert.equal(fs.readFileSync(path.join(destDir, 'file.txt'), 'utf8'), 'content');
+
+    tl.rmRF(srcDir);
+    tl.rmRF(destDir);
+    done();
+  });
 });


nielswitte:
> @nielswitte please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree

nielswitte:
Fixed in #1177 

Closing PR

PR #1177, opened 2026-06-05
Fix glob pattern handling in cp to prevent infinite recursion and support full-path patterns

Deekshitha2812:
### Problem

 The `cp` function's glob handling has issues:

1. **Infinite recursion and broken full-path globs due to dirname/basename split**: The glob detection regex checked the entire source path for special characters like `@`, `[`, `]`, `*`, `?`, but expansion used only `path.basename(source)` via `findMatch(path.dirname(source), [path.basename(source)])`. This caused two problems:
   - For literal paths containing special characters in directory names (e.g. `dir@test/file.txt`, `[Test] dir/file.txt`), the path was misidentified as a glob pattern. Since only the basename was expanded, `findMatch` returned the same literal path, and `cp` recursed on it endlessly until stack overflow.
   - For real full-path globs like `src/**/*.txt` or `packages/*/dist/output.js`, the glob characters in directory segments were split into `path.dirname` which is supposed to be a literal root — so the pattern semantics were lost.

2. **Bare `!` and `#`**: Leading `!` is an exclude operator in `findMatch` that removes files from a prior include set. Since `cp` takes a single source — not a pattern list — there is no prior include set for `!` to act on. Leading `#` would cause `findMatch` to silently skip the source as a comment. Neither behavior is appropriate for `cp`.

### Changes

- Removed `^[#!]` from the glob detection regex. The extglob operator `!(...)` is still correctly detected via `[@+!]\(`.
- Changed `cp` to pass the **full source pattern** to `findMatch` instead of splitting it into `path.dirname`/`path.basename`. This preserves glob semantics across directory segments and eliminates the self-referencing recursion.
- Added `{ nonegate: true, nocomment: true }` when calling `findMatch` from `cp` to prevent leading `!` and `#` from being consumed by `findMatch`'s list-level pre-processing.
- Added a shell-style fallback: when glob expansion returns zero matches, `cp` falls through to treat the source as a literal path. This matches how the previous `shell.cp` (shelljs) behaved — try pattern expansion first, and if nothing matches, copy the original path literally.

### Tests added

- **`cp falls back to literal copy when glob pattern returns no results`**: Copies a file from a directory with `[` and `]` in its name without infinite recursion or failure, verifying the literal fallback path.
- **`cp with recursive full-path glob pattern`**: Uses a pattern like `src/**/*.txt` across a nested directory structure and verifies that only matching files are copied while non-matching files are excluded.

Related issue - https://github.com/microsoft/azure-pipelines-task-lib/issues/1166

### DIFF
--- a/node/task.ts
+++ b/node/task.ts
@@ -1232,19 +1232,18 @@ export function cp(sourceOrOptions: unknown, destinationOrSource: string, option
         if (!fs.existsSync(destination) && !force) {
             throw new Error(loc('LIB_PathNotFound', 'cp', destination));
         }
-        const isPattern = /[*?{\[]/.test(source) || /^[#!]/.test(source) || /[@+!]\(/.test(source);
+        const isPattern = /[*?{\[]/.test(source) || /[@+!]\(/.test(source);
         if (isPattern) {
-            let sourcesToProcess: string[] = [];
-            let sourceDir = path.dirname(source);
-            sourceDir = sourceDir == '.' ? path.resolve() : sourceDir;
-            sourcesToProcess = findMatch(sourceDir, [path.basename(source)]);
-            if (sourcesToProcess.length === 0) {
-                debug(`No matches found for pattern: ${source}`);
+           const defaultRoot = getVariable('system.defaultWorkingDirectory') || process.cwd();
+            const sourcesToProcess = findMatch(defaultRoot, [source], undefined, <MatchOptions>{nonegate: true, nocomment: true});
+            if (sourcesToProcess.length > 0) {
+                for (const src of sourcesToProcess) {
+                    cp(src, destination, options as CopyOptionsVariants, continueOnError, retryCount);
+                }
+                return;
+            } else {
+                debug(`No matches found for the pattern: ${source}. Fallback to check for the literal path.`);
             }
-            for (const src of sourcesToProcess) {
-                cp(src, destination, options as CopyOptionsVariants, continueOnError, retryCount);
-            }
-            return;
         }
         var lstatSource = fs.lstatSync(source);
 
--- a/node/task.ts
+++ b/node/task.ts
@@ -1234,7 +1234,7 @@ export function cp(sourceOrOptions: unknown, destinationOrSource: string, option
         }
         const isPattern = /[*?{\[]/.test(source) || /[@+!]\(/.test(source);
         if (isPattern) {
-           const defaultRoot = getVariable('system.defaultWorkingDirectory') || process.cwd();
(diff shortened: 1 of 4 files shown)


node/task.ts
```diff
@@ -1232,19 +1232,18 @@ export function cp(sourceOrOptions: unknown, destinationOrSource: string, option
...
-            let sourcesToProcess: string[] = [];
-            let sourceDir = path.dirname(source);
-            sourceDir = sourceDir == '.' ? path.resolve() : sourceDir;
-            sourcesToProcess = findMatch(sourceDir, [path.basename(source)]);
-            if (sourcesToProcess.length === 0) {
-                debug(`No matches found for pattern: ${source}`);
-            }
-            for (const src of sourcesToProcess) {
-                cp(src, destination, options as CopyOptionsVariants, continueOnError, retryCount);
+            const defaultRoot = getVariable('system.defaultWorkingDirectory') || process.cwd();
```
dassayantan24, line 1237: defaultWorkingDirectory and sources dir may differ. Is this intentional?

node/task.ts
```diff
@@ -1232,19 +1232,18 @@ export function cp(sourceOrOptions: unknown, destinationOrSource: string, option
         if (!fs.existsSync(destination) && !force) {
             throw new Error(loc('LIB_PathNotFound', 'cp', destination));
         }
-        const isPattern = /[*?{\[]/.test(source) || /^[#!]/.test(source) || /[@+!]\(/.test(source);
+        const isPattern = /[*?{\[]/.test(source) || /[@+!]\(/.test(source);
```
dassayantan24, line 1235: Can we please check if this is backward compatible?

Deekshitha2812: Previously, leading `#` was silently ignored as it was treated as a comment, and leading `!` produced no results since there were no prior files included for it to exclude from. The behavior now is similar — both still yield no matching results from glob expansion. The only difference is that the literal path is now checked as a fallback, so if the file actually exists on disk, it gets copied instead of being silently skipped.

node/task.ts
Deekshitha2812: Yes. `defaultRoot` only affects relative patterns — absolute paths (the common case for `cp`) ignore it entirely. This is consistent with how `findMatch` resolves its own default root internally(line [2189](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/task.ts#L2178)). `findMatch` falls back to `system.defaultWorkingDirectory` first, and then to `process.cwd()` if unset.

tarunramsinghani (approved):

dassayantan24 (approved):

PR #1180, opened 2026-06-25
Replace vulnerable UUID package with internal Node.js crypto for key generation

ivanduplenskikh:
**Summary**

 Replaces the deprecated and vulnerable **uuid** package with Node.js's built-in `crypto.randomUUID()` for generating the Vault encryption key.
This removes a third-party dependency and sources the key from a cryptographically secure generator.

 **What changed**

 •  **node/vault.ts**: Removed `require('uuid/v4').genKey()` now calls the already-imported `crypto.randomUUID()` to produce the **.taskkey** value. Minor style cleanup (semicolons, brace formatting).
 •  **node/package.json** and **node/package-lock.json**: Dropped the **uuid^3.0.1** dependency.

 **Why**

  **uuid@3.4.0** is deprecated and vulnerable - older versions may fall back to Math.random(),  which is not cryptographically secure (details). Because the Vault key seeds AES-256-CTR encryption of secrets, it should come from a CSPRNG. `crypto.randomUUID()` is built into Node.js, cryptographically strong, and eliminates the external package.

 **Testing**

 • Existing Vault unit tests:  node/test/vaulttests.ts .
 • Build and run the Node test suite.

 **Risks and rollback**

 •  **crypto.randomUUID()** requires Node.js ≥ **14.17** - confirm all supported runtimes meet this minimum.
 • The key is hashed with **SHA-256** before use and regenerated each run, so the change in key value/format is backward compatible.
 • Rollback: revert the single commit to restore the **uuid** dependency.

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -14,8 +14,7 @@
         "nodejs-file-downloader": "^4.11.1",
         "q": "^1.5.1",
         "semver": "^5.7.2",
-        "shelljs": "^0.10.0",
-        "uuid": "^3.0.1"
+        "shelljs": "^0.10.0"
       },
       "devDependencies": {
         "@types/minimatch": "3.0.3",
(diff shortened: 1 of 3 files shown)


tarunramsinghani (approved):

aidan-doherty:
Just FYI, this change seems to have stopped one of our build servers from working, with the error "crypto.randomUUID is not a function".

We didn’t have this installed on the server in the first place, so I assume it’s a dependency that was previously included in the pipeline but hasn’t been automatically updated alongside this fix.

ivanduplenskikh:
@aidan-doherty, what's your setup? Is it on-premise or Azure DevOps Cloud?

The possible issue is your agent's using Node.js 10, because [the method was added in Node.js v14/15](https://nodejs.org/api/crypto.html#cryptorandomuuidoptions).

I'd suggest upgrading the Node.js version, since [Node.js v10 has security issues](https://nodejs.org/en/about/eol), is EOL, and [partially out-of-support from Azure Dev Ops perspective in 2022](https://learn.microsoft.com/en-us/azure/devops/release-notes/roadmap/2022/no-node-6-on-hosted-agents) and will be completely deprecated [in November 2026](https://learn.microsoft.com/en-us/azure/devops/release-notes/roadmap/2026/retire-node-6)

aidan-doherty:
> Just FYI, this change seems to have stopped one of our build servers from working, with the error "crypto.randomUUID is not a function".
> 
> We didn’t have this installed on the server in the first place, so I assume it’s a dependency that was previously included in the pipeline but hasn’t been automatically updated alongside this fix.


The agent on this server clearly hasn't been updated in a while as it's on v2 im sure updating it will fix the issue.

ivanduplenskikh:
@aidan-doherty, [the v2 agent](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.220.0) was released in 2023 and works on .NET Core 3.1 that is EOL. In this case, I would strongly recommend upgrading

aidan-doherty:
@ivanduplenskikh thanks Ivan, upgraded now and all is working as expected 👍

PR #1183, opened 2026-07-20
Update vulnerability adm zip to v0.6.0

ivanduplenskikh:
## Summary

Resolves an open high-severity `adm-zip` vulnerability

## What changed

**Dependency updates** (`node/package.json`, `node/package-lock.json`)
- `adm-zip`: `^0.5.10` → `^0.6.0`

## Why

- `adm-zip < 0.6.0` has an open high-severity advisory — a crafted ZIP file can trigger a ~4GB memory allocation (denial of service): [GHSA-xcpc-8h2w-3j85](https://github.com/advisories/GHSA-xcpc-8h2w-3j85) / CVE-2026-39244.

## Testing

- `npm run build` succeeds
- `npm test` succeeds

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-task-lib",
-  "version": "5.277.0",
+  "version": "5.278.0",
   "description": "Azure Pipelines Task SDK",
   "main": "./task.js",
   "typings": "./task.d.ts",
@@ -27,7 +27,7 @@
   },
   "homepage": "https://github.com/Microsoft/azure-pipelines-task-lib",
   "dependencies": {
-    "adm-zip": "^0.5.10",
+    "adm-zip": "^0.6.0",
     "minimatch": "^3.1.5",
     "nodejs-file-downloader": "^4.11.1",
     "q": "^1.5.1",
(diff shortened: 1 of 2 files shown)


node/package.json
tarunramsinghani: We will have to update the package version as well..

node/buildutils.js
```diff
@@ -148,6 +147,13 @@ var downloadArchiveAsync = async function (url, fileName) {
 };
 
 
+var extractZipToDir = function (archivePath, targetPath) {
```
tarunramsinghani: I understand the goal of new function was to add tests, but we do not have logic here to test, just a function call, do we really need to test the functionality of adm-zip ? we are not also just testing the behaviour we depend on that we think might break us in future. Not sure why we need the tests here...

ivanduplenskikh: My intention was to make sure that adm-zip works correctly after version bumping within our azure-pipelines-task-lib module which is used across [hundreds of packages](https://github.com/search?q=path%3Apackage.json+%22azure-pipelines-task-lib%22+owner%3Amicrosoft&type=code)

node/package.json
ivanduplenskikh: Updated

tarunramsinghani (approved):

PR #1205, opened 2026-09-04
Users/shaparashar/vso command injection

ShauryaP1444:
## Summary

Adds filtering for external output before it is displayed in Azure Pipelines task logs.

- Neutralizes unapproved `##vso[` markers in external output.
- Supports an explicit positive allowlist for approved VSO commands.
- Integrates filtering with ToolRunner command-line, `stdout`, `stderr`, piping, `exec`, and `execSync` display paths.
- Preserves raw ToolRunner events and bytes piped between child processes.
- Uses the correct `end` event to detect child `stdout` completion.

### Automated Tests

All tests included in this PR pass successfully.

### End-to-End Test

Validated the changes using the `cURLUploader` task and a controlled HTTP server response containing:

```text
##vso[task.complete ...]
```

The response was displayed in the pipeline log as:

```text
##_vso[task.complete ...]
```

The remote command was successfully neutralized and was not processed by the Azure Pipelines agent. Trusted commands generated by the task itself continued to work as expected.


node/externaloutput.ts
```diff
@@ -0,0 +1,328 @@
...
+            // Allowlisted markers pass unchanged; the header/data after them flow through as
+            // ordinary bytes and any later markers are evaluated independently.
+            out.push(name && this.allowed.has(name) ? MARKER : NEUTRALIZED);
+            pos = afterMarker;
+        }
+
+        if (out.length === 0) {
+            return EMPTY;
+        }
+        return out.length === 1 ? out[0] : Buffer.concat(out);
```
tarunramsinghani: **This fast path can hand a module-level constant back to the caller.**

`out.push(NEUTRALIZED)` / `out.push(MARKER)` push the shared `Buffer` constants themselves. `Buffer.concat` allocates a copy, so the multi-fragment path is fine — but `out.length === 1` returns the element by reference. When a chunk filters down to exactly one marker, the caller receives the module-level constant rather than a buffer it owns.

Two ways to land on it:

```js
filterExternalOutput('##vso[', opts) === filterExternalOutput('##vso[', opts)  // true

const f = new MarkerFilter(false, new Set());
f.push(Buffer.from('##v'));   // partial marker retained in `pending`
f.push(Buffer.from('so['));   // -> out = [NEUTRALIZED], constant returned
```

The result looks like an ordinary per-call buffer, so a caller has no reason to treat it as read-only. If anything writes into it, `NEUTRALIZED` is rewritten for the lifetime of the process and every later filter emits the corrupted bytes:

```js
filterExternalOutput('##vso[', opts).write('##_zzz[');

// afterwards, from any filter instance / ExternalOutputStream / createFilteredWriter:
filterExternalOutput('##vso[task.setvariable variable=x]1', opts)
// -> '##_zzz[task.setvariable variable=x]1'
```

Verified against a local build of this branch. It's cross-instance and sticky — a freshly constructed `MarkerFilter` is affected too — and it fails silently, so the output corruption is very hard to trace back here. Granted, a caller has to write into the returned buffer for this to bite and that isn't a common pattern, but it's cheap to rule out.

Suggest keeping the zero-copy fast path (the `push(input) === input` test at `externaloutputtests.ts:52` relies on it) and copying only the constants:

```suggestion
        if (out.length === 1) {
            const only = out[0];
            // Never hand back module-level state: a caller mutating the returned
            // buffer would corrupt every later filter in the process.
            return only === NEUTRALIZED || only === MARKER ? Buffer.from(only) : only;
        }
        return Buffer.concat(out);
```

Markers are rare relative to ordinary output, so the extra allocation is negligible. Probably worth a regression test asserting `filterExternalOutput('##vso[', o) !== filterExternalOutput('##vso[', o)`.

ShauryaP1444: Thanks, makes sense. I'll apply the suggested changes. I'll add the corresponding test as well. 

I was also considering copying whenever adding MARKER or NEUTRALIZED to out, as that would ensure our internal buffer references never enter out buffer. It just seems more consistent and simpler to understand for me. The performance tradeoff should be negligible as the only redundacy is when Buffer.concat() copies an already-copied marker into the final output buffer.

node/toolrunner.ts
```diff
@@ -40,6 +41,14 @@ export interface IExecSyncOptions {
...
     /** optional. Run command inside of the shell.  Defaults to false. */
     shell?: boolean;
+
+    /**
+     * Optional. When set, the tool's *displayed* output (the command line, stdout and stderr
+     * copies written to the log) is passed through the external-output marker filter so that
+     * "##vso[" commands emitted by the child process are neutralized. Raw stdout/stderr/stdline/
+     * errline events and bytes piped to another tool are NOT affected. See ExternalOutputOptions.
+     */
+    externalOutput?: eom.ExternalOutputOptions;
```
tarunramsinghani, line 51: we will need to keep this interface same in mock-toolrunner.ts file as well ?

ShauryaP1444: Thanks. We didn't have any mock tests covering externalOutput, so I missed keeping the mock interface in sync. I've pushed a commit to sync the interfaces and add 2 mock tests. I ran all the mock tests with the updated interface and all tests pass.

sanjuyadav24 (approved):

.taskkey
```diff
@@ -0,0 +1 @@
+838df665-f3ba-4ca6-96de-5f4d54d49ee6
```
tarunramsinghani: cleanup this file and add the entry to do gitignore this pattern of file . seems this is a generated file by tests.

tarunramsinghani (approved):

.taskkey
ShauryaP1444: Thanks. Added to gitignore.

PR #1217, opened 2026-09-14
added unzip functionality

sanjays-ms:
[AB#2462925](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2462925)
[AB#2454589](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2454589)

The extract-zip package used in the pipeline tasks is marked as vulnerable
The package was last updated 6 years ago: https://www.npmjs.com/package/extract-zip

So we are adding a new implementation to unzip files using the yauzl library which is the internal package being used in extract-zip
https://www.npmjs.com/package/yauzl

extract-zip is a [single file project](https://github.com/max-mapper/extract-zip/blob/master/index.js) so this implementation is a copy of that but there are some changes made

The vulnerability states
```
extract-zip through 2.0.1 containment-checks only the parent directory of each archive entry and never the entry's own final path component, so an archive containing two entries with identical names - a symlink whose target is outside the destination, followed by a regular file - writes through the planted symlink and yields an arbitrary file write outside the destination directory

```

Leading to things like
- put a file outside the destination folder
- contain a symlink pointing to an external (potentially dangerous binary) or writing or updating external locations outside of the destination folder
- overwrite an existing symlink

The copied over implementation addresses these issues.

Testing done
extract-zip is being used in 6 tasks
https://github.com/microsoft/azure-pipelines-tasks/pull/22429 - initial testing shows the tasks and existing test cases pass without any issues

However additional test cases are being added in PRs
https://github.com/microsoft/azure-pipelines-canary/pull/1413
https://dev.azure.com/canarytest/PipelineTasks/_git/PipelineTasks/pullrequest/221
to help with testing more of the unzip work flow.

Risk Assessment - Low - this is a new functionality added doesn't update any existing functionality
But the tasks being updated will need to be tested end to end and that will be higher risk.


sylvesterkaczmarek (commented):
Symlink entries are read fully into memory before validation. A zip can mark a very large compressed entry as a symlink and make readEntry() buffer the whole uncompressed payload. Since a link target should only be path-sized, can we cap the symlink entry size before reading it?

PR #1219, opened 2026-09-15
unified logger

ShauryaP1444:
Implemented a dedicated logging api which takes in the type of log - 'debug', 'error', 'warn', filters the log line and uses the corresponding task-lib logger.


node/externaloutput.ts
```diff
@@ -286,11 +286,11 @@ export function createExternalOutputStream(options: ExternalOutputOptions): Exte
  */
 export function writeExternalOutput(data: string | Buffer, options: ExternalOutputOptions): void {
```
tarunramsinghani: We dont need this any more ?

node/externaloutput.ts
```diff
@@ -286,11 +286,11 @@ export function createExternalOutputStream(options: ExternalOutputOptions): Exte
...
     const destination = options.destination || process.stdout;
-    destination.write(filterExternalOutput(data, options));
+    destination.write(_filterExternalOutput(data, options));
 }
 
-/** Filters one complete value and returns its bytes without writing them. */
-export function filterExternalOutput(data: string | Buffer, options: ExternalOutputOptions): Buffer {
+/** Internal helper for consumers that must pass filtered text to another task-lib API. */
+export function _filterExternalOutput(data: string | Buffer, options: ExternalOutputOptions): Buffer {
```
tarunramsinghani: Do we need to export this ?

node/task.ts
```diff
@@ -2835,14 +2847,27 @@ export function writeExternalOutput(data: string | Buffer, options: eom.External
...
- * @param data      The external output to filter.
- * @param options   External output options. See ExternalOutputOptions.
- * @returns         The filtered output.
+ * @param message   The external log message.
+ * @param options   External output and log options.
+ * @returns         void
  */
-export function filterExternalOutput(data: string | Buffer, options: eom.ExternalOutputOptions): Buffer {
-    return eom.filterExternalOutput(data, options);
+export function logExternalOutput(message: string, options: ExternalOutputLogOptions): void {
```
tarunramsinghani: can we also export friendly funcntions for debug, warning,error and normal output (instead of console.out as wrappers on top

node/task.ts
```diff
@@ -2835,14 +2847,27 @@ export function writeExternalOutput(data: string | Buffer, options: eom.External
...
+    const filtered = eom._filterExternalOutput(message, options).toString('utf8');
+
+    switch (options.type) {
+        case 'debug':
```
tarunramsinghani: we are missing normal out option?

node/task.ts
```diff
@@ -2804,6 +2804,18 @@ exports.ToolRunner = trm.ToolRunner;
  */
 export type ExternalOutputOptions = eom.ExternalOutputOptions;
 
+/** The existing task-lib logging path to use after external output is filtered. */
+export type ExternalOutputLogType = 'debug' | 'warning' | 'error';
```
tarunramsinghani: We are missing normal output for console.write or console.out?

node/externaloutput.ts
ShauryaP1444: I was thinking of keeping it for raw writes but, we can integrate it with the unified logger.

node/externaloutput.ts
ShauryaP1444: ohh right, I'll remove this.

node/task.ts
ShauryaP1444: I was thinking of a unified API like:

`tl.logExternalOutput(message, {
    source: 'remote',
    type: 'raw' // 'raw' | 'debug' | 'warning' | 'error'
});`

and use it for the 4 log types with options.type. 

Do we also want wrappers such as debugExternalOutput, writeExternalOutput, etc, or is the single logExternalOutput API with a type option preferable?

ShauryaP1444: Added the friendly functions wrapping logExternalOutput.

node/externaloutput.ts
```diff
@@ -331,3 +322,62 @@ export function createFilteredWriter(options: ExternalOutputOptions, destination
...
+    auditAction?: im.IssueAuditAction;
+};
+
+/** Filters external output before writing it or submitting it through a task-lib logger. */
+function logExternalOutput(message: string | Buffer, options: ExternalOutputLogOptions): void {
+    if (options.type === 'raw') {
+        const destination = options.destination || process.stdout;
+        destination.write(filterExternalOutput(message, options));
+        return;
+    }
```
tarunramsinghani: WHy Raw is handled with if and other optios with switch case ?

ShauryaP1444: Yes, it is inconsistent, I'll move it inside switch case.

tarunramsinghani (approved):

